PR #35, opened 2024-06-11
Implement typeshare

hatchan:
This is intended as a simple proof of concept. Most likely we will not go with Typeshare, but we can still base a different PR on this since this does contain the xtask implementation.


hatchan:
This is superseded by #42

PR #38, opened 2024-06-12
Set up mlflow

actualwitch:
this pr adds a jupyter notebook for easy evaluation of embedding models, mlflow ui for it with all state stored inside the repo (probably not the best idea, but fixing it would required additional infra) and embedding cache in safetensors format


flenter:
Can you add a readme?

brettimus:
closing for now, we can reopen if we decide to evaluate a smaller fine tuned model for req generation

PR #42, opened 2024-06-17
Json Schema / Zod Schema xtask

stephlow:
Introduces json / zod schema generation for Mizu using xtask. Running `cargo xtask generate-schemas` will generate a `frontend/src/schemas.ts` which will export structs / enums / etc defined in Rust to a json schema and convert it into a Zod schema and it's inferred TypeScript type 

Fixes FP-3771

This currently generates `frontend/src/schemas.ts`:

```ts
import { z } from "zod";

export const ClientMessageSchema = z
  .object({
    messageId: z
      .string()
      .describe(
        "A unique identifier for this message. This will be used by certain server messages to refer back to this message, such as Ack or Error.",
      ),
  })
  .and(z.object({ type: z.literal("debug") }))
  .describe("Messages that are send from the client to the server.");

export type ClientMessage = z.infer<typeof ClientMessageSchema>;

export const RequestSchema = z
  .object({
    body: z.union([z.string(), z.null()]).optional(),
    headers: z.record(z.string()),
    id: z.number().int().gte(0),
    method: z.string(),
    url: z.string(),
  })
  .describe("A request that has been captured by fpx.");

export type Request = z.infer<typeof RequestSchema>;

export const RequestAddedSchema = z.object({
  inspector_id: z
    .union([
      z
        .number()
        .int()
        .describe(
          "The id of the inspector that was associated with the request. This is null in the case where the request was send to `/api/inspect`.",
        ),
      z
        .null()
        .describe(
          "The id of the inspector that was associated with the request. This is null in the case where the request was send to `/api/inspect`.",
        ),
    ])
    .describe(
      "The id of the inspector that was associated with the request. This is null in the case where the request was send to `/api/inspect`.",
    )
    .optional(),
  request_id: z
    .number()
    .int()
    .describe("The id of the request that has been captured."),
});

export type RequestAdded = z.infer<typeof RequestAddedSchema>;

export const ServerMessageSchema = z
  .object({
    messageId: z
      .union([
        z
          .string()
          .describe(
            "If this is a response to a client message, then this field contains the same message id. Otherwise it will be [`None`].",
          ),
        z
          .null()
          .describe(
            "If this is a response to a client message, then this field contains the same message id. Otherwise it will be [`None`].",
          ),
      ])
      .describe(
        "If this is a response to a client message, then this field contains the same message id. Otherwise it will be [`None`].",
      )
      .optional(),
  })
  .and(
    z.any().superRefine((x, ctx) => {
      const schemas = [
        z
          .object({ type: z.literal("ack") })
          .describe(
            "A message was received and processed successfully. See the outer message for the message id.",
          ),
        z
          .object({ details: z.any(), type: z.literal("error") })
          .describe(
            "An error occurred on the server. This could be caused by a message or could be caused by something else. See the outer message for the message id.",
          ),
        z
          .object({ details: z.any(), type: z.literal("request_added") })
          .describe(
            "A request has been captured. It contains a reference to the request id and optionally a reference to the inspector id.",
          ),
      ];
      const errors = schemas.reduce<z.ZodError[]>(
        (errors, schema) =>
          ((result) => (result.error ? [...errors, result.error] : errors))(
            schema.safeParse(x),
          ),
        [],
      );
      if (schemas.length - errors.length !== 1) {
        ctx.addIssue({
          path: ctx.path,
          code: "invalid_union",
          unionErrors: errors,
          message: "Invalid input: Should pass single schema",
        });
      }
    }),
  )
  .describe("Messages that are send from the server to the client.");

export type ServerMessage = z.infer<typeof ServerMessageSchema>;

```


.gitignore
```diff
@@ -1,3 +1,5 @@
+frontend/src/schemas.ts
```
hatchan: I think we should still include these in the repo. It will ensure that the developers need to run less command when getting started.

We can prevent any issues with not updating this file by adding a step in the CI which will call this xtasks command, and if the file has changed than it will block the build.

fpx/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod schemas;
```
hatchan: I think that models might be a better name.

fpx/src/schemas.rs
```diff
@@ -106,18 +108,18 @@ impl ClientMessage {
...
     Debug,
 }
 
-#[derive(Clone, Serialize, Deserialize)]
+#[derive(JsonSchema, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct RequestAdded {
     /// The id of the request that has been captured.
-    request_id: i64,
+    request_id: u32,
```
stephlow: This PR is based of Benno's earlier work, in [his PR](https://github.com/fiberplane/fpx/pull/35) he mentioned this:
```
This is changed to accommodate with the lack of 64 bit numbers in Javascript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

If we are sticking with Typeshare, then there is a special type defined in there, which we can consider using.
```

flenter (approved):

hatchan (approved):

PR #43, opened 2024-06-18
Implement initial PoC for ApiServerError

hatchan:
Initial overhaul of errors in fpx:
- Returning ApiServerError from a handler 
- Add DbError abstraction for database error (mimics the one from fp/api)
- Add ApiClientError for the API client

Add simple ApiClient

Change casing, tagging for all API models. 

Resolves: FP-3786


fpx/src/api/client.rs
```diff
@@ -0,0 +1,65 @@
...
+    client: reqwest::Client,
+    base_url: Url,
+}
+
+impl ApiClient {
+    pub fn new(client: reqwest::Client, base_url: Url) -> Self {
+        Self { client, base_url }
+    }
+
+    pub async fn do_req<T, E>(&self, method: Method, path: String) -> Result<T, ApiClientError<E>>
```
mellowagain: can we have `path` as `impl Into<String>` to allow passing `&str`?

mellowagain (approved):

fpx/src/api/client.rs
hatchan: Good point. I actually made it a `impl AsRef<str>` as `join` only requires a `&str`.

stephlow (approved):

fpx/src/api/client.rs
```diff
@@ -11,11 +14,18 @@ pub struct ApiClient {
 }
 
 impl ApiClient {
-    pub fn new(client: reqwest::Client, base_url: Url) -> Self {
+    /// Create a new ApiClient with a default reqwest::Client.
+    pub fn new(base_url: Url) -> Self {
+        let client = reqwest::Client::new();
```
mellowagain: do we wanna set a user agent htat identifies us?

mellowagain (approved):

fpx/src/api/client.rs
hatchan: Good idea

PR #51, opened 2024-06-21
FP-3821: Implement http client for requestor

stephlow:
Allows you to execute a request through requestor, which inserts it into the db and returns the response. Includes generated zod schemas to use in the frontend


fpx/src/api/handlers/requestor.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    for (key, val) in headers.iter() {
+        if let Ok(header_name) = HeaderName::from_str(key.as_str()) {
+            if let Ok(header_value) = HeaderValue::from_bytes(val.as_bytes()) {
+                header_map.insert(header_name, header_value);
+            }
+        }
+    }
+
+    reqwest::Client::new()
```
hatchan: We can initialize a client in as a static and call that. That way we are able to reuse the same client and thus benefit from keep alives and the likes.

hatchan: And also make it lazy so it doesn't get initialized if it isn't being used: 

Something like this: https://github.com/fiberplane/monofiber/blob/3a5a25fb290e16725925ea59b4f90c343fbcbc57/api/src/webhooks/mod.rs#L36-L53

fpx/src/models.rs
```diff
@@ -170,3 +176,84 @@ impl Request {
...
+            status,
+            url,
+            headers,
+            body: Some(body),
+        }
+    }
+}
+
+/// The payload that describes the request that Requestor has to execute
+// TODO: This should probably not be in models?
```
hatchan: This is fine in the models.rs

hatchan (approved):

PR #52, opened 2024-06-24
Log request body

flenter:
And improve rendering of stacktraces

Example of a log message rendering additional data:
image

I've also tweaked how we render logs with different parameters. From top to bottom:
* console logging a the request body as an object
* console logging the request body as a string
* console logging (undefined) (which is rendered as null because json) (The usecase is that sometimes you want to log the return value of something directly and it might be just null/undefined). I've added a TODO for better handling of undefined, but doing that requires a bigger change.
image

Sending (and showing) the body of the request (see bottom of request):
image

### DIFF
--- a/api/src/routes/logs.ts
+++ b/api/src/routes/logs.ts
@@ -144,15 +144,18 @@ async function transformStack(stack: string) {
 
   const parsedLines = await Promise.all(
     lines.map(async (line) => {
-      const regex =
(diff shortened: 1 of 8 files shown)


brettimus:
lgtm — will wait for lau to review too so he gets a sense of what's going on in this PR relative to his

brettimus (approved):

flenter:
I've squashed all commits

laulauland (approved):

PR #54, opened 2024-06-26
Updated request detail page

laulauland:
Resolves FP-3823

Revisiting the request detail view with more expansive and expanded defaults, minimap, and other things

https://github.com/fiberplane/fpx/assets/17225208/568f681f-625b-43ed-9142-fbd06bef2671


frontend/src/pages/RequestDetailsPage/TextJsonViewer.tsx
```diff
@@ -0,0 +1,23 @@
...
+    nordTheme["--w-rjv-background-color"] = "transparent";
+    return (
+      <JsonView
+        value={json}
+        style={nordTheme}
+        displayDataTypes={false}
+        collapsed={collapsed}
+      />
+    );
+  } catch (error) {
```
brettimus, line 20: note to self: test this with a text response

brettimus:
A few notes:

- I modified the layout/grid and made things a little more responsive
- Spacing is a little funky, but we can fix that later
- Should test with _a lot_ of events to make sure overflow doesn't get weird in the minimap
- Minimap gets confused quite easily, probably quite a few edge cases to work through

brettimus (approved):

PR #55, opened 2024-06-26
add canned request saving and listing

mellowagain:


fpx/src/api.rs
```diff
@@ -12,56 +13,72 @@ pub mod handlers;
...
+pub struct FpxDirectoryPath(pub PathBuf);
+
 #[derive(Clone)]
 pub struct ApiState {
     /// The base url on which this server is running. Override this when you
     /// are running this behind a reverse proxy.
     base_url: Url,
 
+    /// the location of the fpx directory
+    fpx_directory: FpxDirectoryPath,
```
hatchan: Maybe we should just wrap the base_url and fpx_directory in a "Config" Struct and wrap that in a Arc. Otherwise we will keep getting more of these wrapping types.

fpx/src/api/handlers/canned_requests.rs
```diff
@@ -0,0 +1,128 @@
...
+use std::collections::BTreeMap;
+use thiserror::Error;
+use tracing::{error, instrument};
+
+#[instrument(skip(fpx_directory_path))]
+pub async fn canned_request_create(
+    State(FpxDirectoryPath(fpx_directory_path)): State<FpxDirectoryPath>,
+    Query(args): Query<CannedRequestCreateArgs>,
+    Json(canned_request): Json<CannedRequest>,
+) -> Result<StatusCode, ApiServerError<CannedRequestCreateError>> {
```
hatchan: I'm not a fan of this pattern. I think we should just stick to the usual pattern where we accept a different payload, than what we return. Than we can move all the query string arguments to json body.

fpx/src/api/handlers/canned_requests.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    name: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, Error)]
+#[serde(tag = "error", content = "details", rename_all = "camelCase")]
+#[non_exhaustive]
+pub enum CannedRequestCreateError {
+    #[error("unknown type, expected `ephemeral`, `personal` or `shared`")]
+    UnknownType,
```
hatchan: `InvalidType` seems better here.

fpx/src/api/handlers/canned_requests.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    #[error("failed to create directory")]
+    DirectoryCreationFailed,
+
+    #[error("failed to serialize")]
+    SerializationFailed,
+
+    #[error("failed to write to file")]
+    FileWriteFailed,
```
hatchan: These seem like very vague errors to me. I think we can map these to simple InternalServerError's and just log it on the server side.

fpx/src/api/handlers/canned_requests.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize, Error)]
+#[serde(tag = "error", content = "details", rename_all = "camelCase")]
+#[non_exhaustive]
+pub enum CannedRequestListError {
+    #[error("failed to convert file name from OsString into &str")]
+    OsStringConversionFailed,
+
+    #[error("file not found")]
```
hatchan:
```suggestion
    #[error("canned request not found")]
```

hatchan (changes requested):

hatchan:
I made some comments, but they ended up on the commit

mellowagain:
re-adding it seemed trickier than it did at first, it adds an implicit bound that the `Internal(anyhow::Error)` variant needed `serde::Deserialize` as well, despite being marked with `#[serde(skip, default)]`. I worked around this by making a newtype which implements these two traits

fpx/src/api/handlers/canned_requests.rs
```diff
@@ -0,0 +1,120 @@
...
+#[derive(Debug, Error, Serialize, Deserialize)]
+#[serde(tag = "error", content = "details", rename_all = "camelCase")]
+#[non_exhaustive]
+#[allow(dead_code)]
+pub enum CannedRequestCreateError {
+    #[error("unknown type, expected `ephemeral`, `personal` or `shared`")]
+    InvalidType,
+
+    #[error("failed to handle request: {0:?}")]
+    Internal(
```
hatchan: I don't think we should have an `Internal` variant, given that we already have the CommonErrors Internal error. 

Regarding the easy propagation using the `?` operator, I think we should investigate where we are returning these anyhow errors and either refactor them to a typed error or just do a `map_error`

I think we might be able to create a `From` for anyhow Error into a `ApiServerError<E>`, which should always result in a Common error. I will investigate this shortly.

hatchan (approved):

fpx/Cargo.toml
```diff
@@ -27,7 +27,7 @@ hex = { version = "0.4" }
 http = { version = "1.1" }
 http-body-util = { version = "0.1" }
 include_dir = { version = "0.7.3" }
-libsql = { version = "0.4", default-features = false, features = ["core", "serde"] }
+libsql = { version = "0.4", default-features = false, features = ["core", "serde", "replication"] }
```
mellowagain: adding this feature is required to make `libsql` compile under windows

PR #58, opened 2024-07-02
Request details tweaks

brettimus:
- Ensure that minimap ids are unique across fetch calls
- Fix KeyValue table styles that i borked last night a bit (key cell should be smaller)
- Fix hidden overflow issues for long minimap TOC (when there are many many events on the request details page). Now, we try to scroll to the entry in the TOC when its outside of the viewport

### DIFF
--- a/frontend/src/pages/RequestDetailsPage/Minimap.tsx
+++ b/frontend/src/pages/RequestDetailsPage/Minimap.tsx
@@ -132,7 +132,7 @@ export function Minimap({ trace }: { trace: MizuTrace | undefined }) {
   }, [toc, handleObserve]);
 
   return (
-    <div className="grid h-[calc(100vh-80px)] overflow-y-auto">
+    <div className="flex flex-col h-[calc(100vh-80px)] overflow-y-auto">
       {toc.length > 0 &&
         toc.map((item, idx) => (
           <div
--- a/frontend/src/pages/RequestDetailsPage/Minimap.tsx
+++ b/frontend/src/pages/RequestDetailsPage/Minimap.tsx
@@ -26,42 +26,42 @@ export function Minimap({ trace }: { trace: MizuTrace | undefined }) {
         const { message } = log;
         if (isMizuRequestStartMessage(message)) {
           return {
-            id: `request-${message.method}-${message.path}`,
+            id: `request-${message.method}-${message.path}-${log.id}`,
             title: "Request",
             method: message.method,
           };
         }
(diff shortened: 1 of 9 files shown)


frontend/src/pages/RequestDetailsPage/Minimap.tsx
```diff
@@ -111,11 +132,12 @@ export function Minimap({ trace }: { trace: MizuTrace | undefined }) {
   }, [toc, handleObserve]);
 
   return (
-    <div className="flex flex-col">
```
laulauland, line 114: any particular reason why you switched to grid here? Just for my understanding

frontend/src/pages/RequestDetailsPage/Minimap.tsx
```diff
@@ -26,42 +26,42 @@ export function Minimap({ trace }: { trace: MizuTrace | undefined }) {
         const { message } = log;
         if (isMizuRequestStartMessage(message)) {
           return {
-            id: `request-${message.method}-${message.path}`,
+            id: `request-${message.method}-${message.path}-${log.id}`,
```
laulauland, line 29: is log.id unique enough? AFAIK it's the trace id right?

laulauland (approved):
couple of questions just for my understanding but LGTM

frontend/src/pages/RequestDetailsPage/Minimap.tsx
laulauland: looks like this is also the reason why the minimap height is a little borked now

frontend/src/pages/RequestDetailsPage/Minimap.tsx
brettimus: log.id is the id of the log record from the libsql db!

frontend/src/pages/RequestDetailsPage/Minimap.tsx
brettimus: yeahhh i only tested with the long list. reverted the change and it fixed!

i used grid because grids-within-grids tend to respect parent height better for scrollability (that was what helped me out in the requestor UI)

but flex-col worked fine here, since the ultimate solution was a hacky-hack that set an explicity height on the minimap container

frontend/src/pages/RequestDetailsPage/Minimap.tsx
laulauland: cool!

PR #60, opened 2024-07-03
Revamp the npx cli user experience, and support safe production deploys

brettimus:
- FP-3839
- FP-3846
- FP-3837

## Demo

https://github.com/fiberplane/fpx/assets/3201327/81125a08-e62f-4afa-b64f-75f517d405d2

## Description

This PR does two things:

1. Modify the cli to _initialize_ fpx. This runs a small wizard that creates a local configuration folder, and a (versioned) configuration file.
2. Update the middleware to only send log/trace/event data to the fpx api when an `FPX_ENDPOINT` is defined in the user's api's environment

(2) Is very simple, just changed a few lines of code in the `client-library`, so the rest of this description will focus on (1)

### Initializing FPX from the CLI

The CLI will create (or read from) `.fpxconfig/fpx.config.v0.json`. This is assumed to be an object whose keys are environment variables. These are considered "User Variables". 

```jsonc
{
  "FPX_SERVICE_TARGET": "8787",
  "FPX_PORT": "8788",
  // This database file path will be set based off of platform (windows vs *nix-like)
  "FPX_DATABASE_URL": "file:.fpxconfig/fpx.db"
}
```

When determining the port on which FPX studio should run, it checks:
- `process.env.FPX_PORT`
- `FPX_PORT` from the User Variables (config file in project)

It falls back to asking the user for the port on which it should run. The default is 8788.

When determining the url for the user's service, it checks:
- `process.env.FPX_SERVICE_TARGET`
- `FPX_SERVICE_TARGET` from the User Variables
-  The `[dev.port]` defined in the `wrangler.toml` file, if one exists

Then, it also silently configures the service name based on what's in wrangler.toml or package.json (if one exists). 

In addition to the above it:

- Checks if the FPX_PORT is available, and suggests a new one if it is not
- Checks if the user's api is running, and reminds them to start it if it is not
- Creates `.dev.vars` if one does not exist (and if we detect a wrangler.toml file)
- Asks permission to modify the env var file to update `FPX_ENDPOINT` accordingly, if it is out of date where we'll serve the studio

## Testing locally

There are instructions in the api readme for this, but @keturiosakys pointed out it's also possible to just run

```sh
npx ../path/to/fpx/api
```
But! You need to build the studio before running that.

```sh
cd /path/to/fpx
# from the project root!
npm run build:fpx-studio
```


api/package.json
```diff
@@ -4,6 +4,9 @@
   "description": "Local development debugging interface for Hono apps",
   "author": "Fiberplane<info@fiberplane.com>",
   "type": "module",
+  "engines": {
```
brettimus, line 7: This is the base requirement for Hono, seems safe to put it here

Also we want esm support for the CLI, which would require > 12

api/src/app.ts
```diff
@@ -2,12 +2,13 @@ import { createClient } from "@libsql/client";
...
 import { Hono } from "hono";
 import { env } from "hono/adapter";
-import { logger } from "hono/logger";
+import { logger as honoLogger } from "hono/logger";
 import type { WebSocket } from "ws";
 
 import { DEFAULT_DATABASE_URL } from "./constants.js";
 import * as schema from "./db/schema.js";
 import type { Bindings, Variables } from "./lib/types.js";
+import logger from "./logger.js";
```
brettimus, line 11: The logger we're using for app logs (to replace console.* in the api) is aware of an env var called FPX_LOG_LEVEL, which allows us to hide some logs from end users of the cli. 

Basically, we don't wanna pollute their terminal with our crazy logs, but if you set `FPX_LOG_LEVEL=debug` locally, it will print all the stuff that's helpful for local debugging

api/bin/cli.js
```diff
@@ -1,9 +1,12 @@
-#!/usr/bin/env node
+#!/usr/bin/env node --experimental-modules
```
brettimus, line 1: we run with --experimental-modules to have backwards compatibility with node 16

api/bin/logger.js
```diff
@@ -0,0 +1,34 @@
+/**
```
brettimus, line 1: copy-pasted-simplified version of the logger added to the api, just fyi :)

PR #63, opened 2024-07-09
Re-discover routes on app update

laulauland:
Resolves FP-3845

Updates the Typescript API to use file watching to restart route probes. This assumes `npx @fiberplane/studio` gets run from the root of the project in question (or that directory is provided over an environment variable).

### DIFF
--- a/api/src/index.node.ts
+++ b/api/src/index.node.ts
@@ -1,12 +1,13 @@
 import fs from "node:fs";
 import { createServer } from "node:http";
+import path from "node:path";
 import { serve } from "@hono/node-server";
 import { config } from "dotenv";
+import { minimatch } from "minimatch";
 import { type WebSocket, WebSocketServer } from "ws";
-
 import { createApp } from "./app.js";
 import { getIgnoredPaths } from "./lib/utils.js";
-import { probeRoutesWithExponentialBackoff } from "./probe-routes.js";
+import { debouncedProbeRoutesWithExponentialBackoff } from "./probe-routes.js";
 import {
   frontendRoutesHandler,
   staticServerMiddleware,
@@ -50,12 +51,38 @@ const watchDir = process.env.FPX_WATCH_DIR ?? process.cwd();
 
 const ignoredPaths = getIgnoredPaths();
 
+debouncedProbeRoutesWithExponentialBackoff(
+  serviceTargetArgument,
+  probeMaxRetries,
+  probeDelay,
+);
+
 fs.watch(watchDir, { recursive: true }, async (eventType, filename) => {
-  if (!filename) return;
-  if (ignoredPaths.includes(filename)) return;
+  if (!filename) {
+    return;
+  }
+  if (ignoredPaths.includes(filename)) {
+    return;
+  }
+  // E.g., ignore everything inside the `.wrangler` directory
+  if (
+    ignoredPaths.some((pattern) => filename.startsWith(`${pattern}${path.sep}`))
+  ) {
+    return;
+  }
+  // E.g., ignore all files with the given name (e.g., `fpx.db`, `fpx.db-journal`)
+  if (ignoredPaths.some((pattern) => path.basename(filename) === pattern)) {
+    return;
+  }
+  // Use minimatch to see if the filename matches any of the ignored patterns
+  // This is because gitignore can have glob patterns, etc
+  if (ignoredPaths.some((pattern) => minimatch(filename, pattern))) {
+    return;
+  }
+
   console.debug(`File ${filename} ${eventType}, sending a new probe`);
 
-  probeRoutesWithExponentialBackoff(
+  debouncedProbeRoutesWithExponentialBackoff(
     serviceTargetArgument,
     probeMaxRetries,
     probeDelay,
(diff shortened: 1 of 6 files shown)


api/src/index.node.ts
```diff
@@ -44,11 +46,22 @@ console.log(`FPX Server is running: http://localhost:${port}`);
...
-  probeMaxRetries,
-  probeDelay,
-);
+const watchDir = process.env.FPX_WATCH_DIR ?? process.cwd();
+
+const ignoredPaths = getIgnoredPaths();
+
+fs.watch(watchDir, { recursive: true }, async (eventType, filename) => {
+  if (!filename) return
+  if (ignoredPaths.includes(filename)) return
```
stephlow: I guess this will still use ignored files if their parent folder is only declared in the ignored path?
Something like `target` where the filename could me `target/file.*`. Maybe we should do glob matching here?

I know there's only hardcoded values right now but this could potentially bite us in the rear later on

stephlow (approved):

api/src/index.node.ts
laulauland: yea I was torn how far to go with it for this initial implementation. To some extent the cost of the probe restarting more often than perhaps it should is minimal so I didn't bother with a more comprehensive ignorePath resolution. At the same time I'm open to making this more robust later on.

stephlow: Yeah would agree, maybe a ticket or a comment should be fine for now :)

brettimus: i ended up implementing glob matching, since we were effectively watching all of `.wrangler`, which got quite noisy and i thiiink was causing some strange errors with wrangler itself?

brettimus:
thing to look out for: i kept seeing this error

```
✘ [ERROR] Error in ProxyController: Error inside ProxyWorker

   Error
      at castErrorCause
  (/Users/brettbeutell/fiber/goose-quotes/node_modules/wrangler/wrangler-dist/cli.js:207265:18)
      at ProxyController.emitErrorEvent
  (/Users/brettbeutell/fiber/goose-quotes/node_modules/wrangler/wrangler-dist/cli.js:208155:16)
      at ProxyController.onProxyWorkerMessage
  (/Users/brettbeutell/fiber/goose-quotes/node_modules/wrangler/wrangler-dist/cli.js:208051:14)
      at PROXY_CONTROLLER
  (/Users/brettbeutell/fiber/goose-quotes/node_modules/wrangler/wrangler-dist/cli.js:207826:20)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async #handleLoopbackCustomService
  (/Users/brettbeutell/fiber/goose-quotes/node_modules/miniflare/dist/src/index.js:8765:24)
      at async #handleLoopback
  (/Users/brettbeutell/fiber/goose-quotes/node_modules/miniflare/dist/src/index.js:8807:20) {
    cause: {
      name: 'Error',
      message: 'Network connection lost.',
      stack: 'Error: Network connection lost.'
    }
  }
```

which i haven't seen before. curious if implementing a filewatcher had some effect on wranglers reloading.

api/src/lib/utils.ts
```diff
@@ -29,3 +33,64 @@ export function errorToJson(error: Error) {
     // Optionally add more properties here if needed
   };
 }
+
+export function shouldIgnoreFile(
```
brettimus, line 37: factored out ignore logic here, but got too lazy to write unit tests... but now i'm writing this i realize i could just write some with cursor

brettimus: okay tested!

api/src/lib/utils.ts
```diff
@@ -29,3 +33,64 @@ export function errorToJson(error: Error) {
...
+    "fpx.db-journal",
+    "mizu.db",
+    "mizu.db-journal",
+    ".fpx",
+    ".fpxconfig",
+    ".swc",
+    ".wrangler",
+  ];
+
+  try {
```
brettimus, line 71: added a try-catch here in case there isn't a .gitignore in the directory

api/src/probe-routes.ts
```diff
@@ -1,8 +1,70 @@
...
+      func(...args);
+    }, wait);
+  };
+}
+
+/**
+ * Since we are calling the route probe inside a file watcher, we should implement
+ * debouncing to avoid spamming the service with requests.
+ *
+ * HACK - Since we are monitoring ts files, we need a short delay to let the code
```
brettimus, line 24: i actually did run into race conditions locally (i think?). adding a delay helped reduce the amount of times that my route changes didn't propagate

api/src/probe-routes.ts
```diff
@@ -1,8 +1,70 @@
...
+ *        for the service recompile.
+ */
+const debouncedProbeRoutesWithExponentialBackoff = debounce(
```
brettimus, line 27: debounced this at first because i realized we were watching `.wrangler` which caused many many probes to go off. after implementing minimatch logic that mitigated the need for this but i kept it anyhow.

brettimus (approved):

brettimus:
gonna merge but left comments for any who are interested in the changes i added

PR #65, opened 2024-07-10
Fix favicons, keyboard hook swap, mod+enter on main

laulauland:
FP-3858

Small PR to ensure the favicons get built and injected into the head of the page.

Also swaps out homebrew keyboard hook for a library one that is more robust.

Finally, it adds a `mod+enter` (`cmd+enter` on mac, `ctrl+enter` elsewhere) keyboard shortcut for submitting requests on the requestor page


frontend/src/pages/RequestorPage/RequestorPage.tsx
```diff
@@ -109,6 +110,16 @@ export const RequestorPage = () => {
     selectedRoute,
   });
 
+  const formRef = useRef<HTMLFormElement>(null);
+
+  useHotkeys("mod+enter", () => {
+    if (formRef.current) {
+      formRef.current.requestSubmit();
```
brettimus: this was a little simpler than dispatching an event manually

frontend/src/pages/RequestorPage/RequestorPage.tsx
```diff
@@ -109,6 +110,16 @@ export const RequestorPage = () => {
...
+    }
+  }, {
+    enableOnFormTags: ["input"],
```
brettimus: need to specifically enable on form tags so that it works when the input is focused

frontend/src/components/ui/DataTable.tsx
```diff
@@ -137,9 +137,9 @@ export function DataTable<TData, TValue>({
     }
   });
 
-  useKeySequence(["j"], handleNextRow);
-  useKeySequence(["k"], handlePrevRow);
-  useKeySequence(["Enter"], handleRowSelect);
+  useHotkeys(["j"], handleNextRow);
+  useHotkeys(["k"], handlePrevRow);
+  useHotkeys(["Enter"], handleRowSelect, { preventDefault: true });
```
brettimus, line 142: this prevents a bug when running with vite (i think) since hitting enter can also trigger a hot reload locally

brettimus (approved):

PR #67, opened 2024-07-11
Tracing UI with hackadoo "Spans"

brettimus:
The goal of this PR is to massage the existing MizuTrace model into something that also contains spans, and can ultimately render a UI like this:

image

Took data modeling inspo from Otel conventions as well as the open PR on the rust side of things to ingest otel traces

***

## Summary

> **TODO**


frontend/src/pages/RequestDetailsPage/RequestDetailsPage.tsx
```diff
@@ -111,6 +112,115 @@ export function RequestDetailsPage() {
...
+    if (!traceV2) {
+      return <EmptyState />;
+    }
+    return (
+      <div
+        className={cn(
+          "h-full",
+          "relative",
+          "overflow-hidden",
+          "overflow-y-scroll",
```
flenter: Why are we using overflow-y-scroll as opposed to overflow-y-auto? Depending if people use a mouse (on a mac) or whatever OS they are using or how it's configured, users might always see scrollbars even if there's nothing to scroll.

flenter (approved):
This looks ok to me. One thing that would be nice to change is to use `overflow-x/y-auto` instead of the `scroll`. Cause if the OS doesn't auto hide scrollbars you otherwise always see them (even when there's nothing to scroll). 

There are some improvements possible like for instance limiting the response that is being displayed atop the request details and making some other sections collapsible
image

frontend/src/pages/RequestDetailsPage/RequestDetailsPage.tsx
brettimus: yup fixed

brettimus:
> There are some improvements possible like for instance limiting the response that is being displayed atop the request details and making some other sections collapsible 

yes yes - very good catch. i just made the summary response section have a max height and be scroll-auto

PR #74, opened 2024-07-15
Update react-query

actualwitch:
we were using old version of react-query, i updated it and made some changes to existing code in accordance with the changes to the package: it uses isPending now instead of isLoading, isError is included in useQuery and query keys are passes as property on the object. note: i did not do comprehensive testing to check that everything works as expected, will do this before merging


stephlow (approved):

frontend/src/pages/RequestorPage/RequestorPage.tsx
```diff
@@ -93,7 +93,7 @@ export const RequestorPage = () => {
     sessionHistory,
   );
 
-  const { mutate: makeRequest, isLoading: isRequestorRequesting } =
+  const { mutate: makeRequest, isPending: isRequestorRequesting } =
```
flenter, line 100: I think we probably want to either use `isLoading` or `isPending` everywhere. Now `useRequestDetails` returns `isLoading`, while it's `isPending` here.

flenter (approved):
one comment, apart from that lgtm

frontend/src/pages/RequestorPage/RequestorPage.tsx
actualwitch: i changed it to pending everywhere i could find so that we are aligned with react-query

PR #80, opened 2024-07-18
Middleware: Fix execution context errors in Node, Deno

brettimus:
**NOTE** - This is branched off of #79 

***

We get execution context errors outside of Cloudflare envs (specifically, Deno, Node).

Until we have middleware for different environments, we should try our hardest not to break outside of Cloudflare.

To that end, this PR adds:

- A polyfill for `executionCtx`, specifically `executionCtx.waitUntil`
- Error handling for our call to the local FPX service, so we don't crash long-running processes when FPX Studio is not running

### DIFF
--- a/client-library/package.json
+++ b/client-library/package.json
@@ -4,7 +4,7 @@
   "author": "Fiberplane<info@fiberplane.com>",
   "type": "module",
   "main": "dist/index.js",
-  "version": "0.2.1-beta.1",
+  "version": "0.2.1-beta.2",
   "dependencies": {
     "@neondatabase/serverless": "^0.9.3"
   },
--- a/client-library/package.json
+++ b/client-library/package.json
@@ -4,7 +4,7 @@
   "author": "Fiberplane<info@fiberplane.com>",
   "type": "module",
   "main": "dist/index.js",
-  "version": "0.2.1-beta.3",
+  "version": "0.2.1-beta.4",
   "dependencies": {
     "@neondatabase/serverless": "^0.9.3"
   },
--- a/client-library/package.json
+++ b/client-library/package.json
@@ -4,7 +4,7 @@
   "author": "Fiberplane<info@fiberplane.com>",
   "type": "module",
   "main": "dist/index.js",
-  "version": "0.2.1-beta.2",
+  "version": "0.2.1-beta.3",
   "dependencies": {
     "@neondatabase/serverless": "^0.9.3"
   },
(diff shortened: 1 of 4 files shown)


brettimus:
~fix tested with `0.2.1-beta.2` and it works! just need to clean up the code~

cleaned up the code, tested with `0.2.1-beta.4` and it works!

client-library/src/honoMiddleware.ts
```diff
@@ -94,7 +96,15 @@ export function createHonoMiddleware<App extends HonoApp>(
 
     const service = env<FpxEnv>(c).FPX_SERVICE_NAME || "unknown";
 
-    const ctx = c.executionCtx;
+    const ctx =
+      getRuntimeKey() === "workerd"
+        ? c.executionCtx
+        : {
+            // HACK - Untested
+            waitUntil: async (p: Promise<unknown>) => {
```
brettimus: NOTE - Should swallow this error somehow, otherwise it will be unhandled and crash the Deno runtime

client-library/src/honoMiddleware.ts
```diff
@@ -94,7 +96,15 @@ export function createHonoMiddleware<App extends HonoApp>(
 
     const service = env<FpxEnv>(c).FPX_SERVICE_NAME || "unknown";
 
-    const ctx = c.executionCtx;
+    const ctx =
+      getRuntimeKey() === "workerd"
```
brettimus: need to lookup if this is true for, e.g., vercel runtime

also - do we want to have separate middleware for different runtimes? might make sense from an observability perspective

client-library/src/index.ts
```diff
@@ -1 +1 @@
-export { createHonoMiddleware } from "./honoMiddleware";
+export { createHonoMiddleware } from "./honoMiddleware.js";
```
brettimus: note - these changes with the `.js` extension are because this is branched off of https://github.com/fiberplane/fpx/pull/79

client-library/package.json
```diff
@@ -4,7 +4,7 @@
   "author": "Fiberplane<info@fiberplane.com>",
   "type": "module",
   "main": "dist/index.js",
-  "version": "0.2.0",
+  "version": "0.2.1-beta.3",
```
brettimus: already published if you want to test: https://www.npmjs.com/package/@fiberplane/hono?activeTab=versions

client-library/package.json
```diff
@@ -4,7 +4,7 @@
   "author": "Fiberplane<info@fiberplane.com>",
   "type": "module",
   "main": "dist/index.js",
-  "version": "0.2.0",
+  "version": "0.2.1-beta.4",
```
brettimus, line 7: published on npm if you want to test it out

flenter (approved):

PR #82, opened 2024-07-20
CLI: Handle Deno project detection better when initializing

brettimus:
Tested with `npx @fiberplane/studio@0.3.0-beta.7`

***

I ran into an issue during the Honcathon, where the FPX cli didn't play nicely with our Deno project.

The issue lies in how we detect the project root. Current CLI only looks for `package.json` or `wrangler.toml`. It does not consider a `deno.json` or `deno.jsonc` file. 

Since I had a rogue `package.json` a few levels above the project root, my `.fpxconfig` ended up being created outside of the project directory.

This PR fixes the issue by looking for the closest matching filepath to the current directory amongst `wrangler.toml`, `package.json`, `deno.json`, and `deno.jsonc`.

***

This also includes a fix that considers `host.docker.internal` the same as `localhost` when detecting whether or not the user's .env definition of `FPX_ENDPOINT` needs to be changed.

During the Honcathon, our project used tilt + docker containers. So, to access fpx, our API needed to specify `host.docker.internal`. Every time I ran the CLI, it asked if i wanted to update the fpx endpoint, since it didn't use localhost.

### DIFF
--- a/api/bin/cli.js
+++ b/api/bin/cli.js
@@ -152,10 +152,7 @@ async function getFpxPort() {
 async function updateEnvFileWithFpxEndpoint(fpxPort) {
   const LOCALHOST_ENDPOINT = `http://localhost:${fpxPort}/v0/logs`;
   const DOCKER_ENDPOINT = `http://host.docker.internal:${fpxPort}/v0/logs`;
-  const expectedFpxEndpoints = [
-    LOCALHOST_ENDPOINT,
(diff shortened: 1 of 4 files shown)


brettimus:
for the record: i've tested on a Deno app and Cloudflare Workers app, both work as expected

stephlow (approved):

PR #83, opened 2024-07-23
Client-library-otel

flenter:
This is the initial implementation of the open telemetry compliant version of the fpx client library.

The api is slightly different. First the current library:

``` ts
import { createHonoMiddleware } from "@fiberplane/hono";

const app = new Hono()
app.use(createHonoMiddleware(app))

app.get("/", (c) => {
  return c.text("Hello Hono!");
});

export default app;
```

New:

``` ts 
import { instrument } from "@fiberplane/hono-otel"

const app = new Hono()

app.get("/", (c) => {
  return c.text("Hello Hono!");
});

export default instrument(app);
```

There's also a measure function which you can use to wrap functions you want to measure.

So:

``` ts
// Simple for testing synchronous javascript execution
const loop = measure("loop", (n: number) => {
  for (let i = 0; i < n; i++) {
    console.log(`Loop iteration: ${i}`);
  }
});
```

Or a function that returns a promise:

``` ts
// Simple function that sleeps for a bit, logs the duration and returns it
const sleep = measure("sleep", (ms: number) => {
  return new Promise<number>((resolve) =>
    setTimeout(() => {
      console.log(`Slept for ${ms}ms`);
      resolve();
    }, ms),
  );
});
```

Instead of passing in a string, you can also pass in an options object, which you can use to do things like setting attributes on the active span and change the spanKind:

``` ts
type MeasureOptions<
  /**
   * Arguments for the function being measured
   */
  ARGS,
  /**
   * The return type of the function being measured
   * (awaited result if the return value is a promise)
   */
  RESULT,
  /**
   * The raw return type of the function being measured
   * (it is used to determine if the onEnd function can be async)
   */
  RAW_RESULT,
> = {
  name: string;
  /**
   * The kind of the span
   */
  spanKind?: SpanKind;
  /**
   * Attributes to be added to the span
   */
  attributes?: Attributes;

  onStart?: (span: Span, args: ARGS) => void;
  /**
   * Allows you to specify a function that will be called when the span ends
   * and will be passed the span & result of the function being measured.
   *
   * This way you can do things like add additional attributes to the span
   */
  onEnd?: (
    span: Span,
    result: RESULT,
  ) => RAW_RESULT extends Promise<unknown> ? Promise<void> | void : void;

  /**
   * Allows you to specify a function that will be called when the span ends
   * with an error and will be passed the (current) span & error that occurred.
   *
   * This way you can do things like add additional attributes to the span
   */
  onError?: (span: Span, error: unknown) => void;
};
```


stephlow (approved):

client-library-otel/src/instrumentation.ts
```diff
@@ -0,0 +1,165 @@
...
+        const originalFetch = value as Hono["fetch"];
+        return async function fetch(
+          request: Request,
+          env: unknown,
+          executionContext: ExecutionContext | undefined,
+        ) {
+          // NOTE - We used to have a handy default for the fpx endpoint, but we need to remove that,
+          //        so that people won't accidentally deploy to production with our middleware and
+          //        start sending data to the default url.
+          const endpoint = (env as Record<string, string | null>).FPX_ENDPOINT;
```
brettimus: might do an optional chain here for safety. i think you're almost always given an object, but there might be certain scenarios (manually invoking fetch in tests) where env is undefined?

client-library-otel/wrangler.toml
```diff
@@ -0,0 +1,3 @@
+name = "hotel"
```
brettimus: can we remove the wrangler.toml file? wasn't sure what this was for per se

client-library-otel/src/utils/wrapper.ts
```diff
@@ -0,0 +1,5 @@
+export function isWrapped(func: unknown) {
+  return (
+    typeof func === "function" && "__wrapped" in func && func.__wrapped === true
```
brettimus, line 3: use a symbol instead of a string to indicate it's wrapped?

client-library-otel/src/utils/request.ts
```diff
@@ -0,0 +1,164 @@
...
+      );
+    } catch {
+      // ignore errors
+    }
+  }
+
+  const headers = response.headers;
+  const responseHeaders = headersToObject(headers);
+  attributes[FPX_RESPONSE_HEADERS_FULL] = JSON.stringify(responseHeaders);
+  for (const [key, value] of Object.entries(responseHeaders)) {
```
brettimus, line 159: 🙏

client-library-otel/src/instrumentation.ts
```diff
@@ -0,0 +1,165 @@
...
+            },
+            originalFetch,
+          );
+
+          const result = await measuredFetch(request, env, proxyExecutionCtx);
+
+          // Make sure all promises are resolved before sending data to the server
+          if (proxyExecutionCtx) {
+            proxyExecutionCtx.waitUntil(
+              Promise.all(promises).finally(() => {
```
brettimus: `Promise.allSettled` instead of `all`? i think `Promise.all` will reject as soon as *any* promise rejects, which means the array of promises will be considered settled and `finally` will execute

`allSettled` makes sure that everything resolves/rejects, and then the `finally` handler executes

client-library-otel/src/instrumentation.ts
```diff
@@ -0,0 +1,165 @@
...
+
+          const serviceName =
+            (env as Record<string, string | null>).FPX_SERVICE_NAME ??
+            "unknown";
+
+          // Patch the related functions to monitor
+          const {
+            monitor: { fetch: monitorFetch, logging: monitorLogging },
+          } = mergeConfigs(defaultConfig, config);
+          monitorLogging && patchConsole();
```
brettimus: nit: i'd say use `if` statements here instead of `&&` (intent is clearer)

client-library-otel/src/instrumentation.ts
```diff
@@ -0,0 +1,165 @@
...
+
+import { measure } from "./measure";
+import { patchConsole, patchFetch, patchWaitUntil } from "./patch";
+import { getRequestAttributes, getResponseAttributes } from "./utils";
+
+type FpxConfig = {
+  monitor: {
+    /** Send data to FPX about each fetch call made during a handler's lifetime */
+    fetch: boolean;
+    // TODO - implement this control/feature
```
brettimus: consider this TODO complete :D

brettimus (approved):
playing with it locally and appears to work!

client-library-otel/wrangler.toml
flenter: It's for the sample app. I use `npm run dev` and have a test app that immediately uses the lib that is located in the`src` folder

client-library-otel/src/instrumentation.ts
flenter: Oh nice one. I didn't know about that one

client-library-otel/.npmignore
```diff
@@ -0,0 +1,56 @@
+node_modules
```
brettimus, line 1: remember to add `wrangler.toml` and `sample/*` here as well

PR #86, opened 2024-07-23
Requestor API routes

stephlow:


frontend/src/schemas.ts
```diff
@@ -58,24 +73,32 @@ export const RequestAddedSchema = z.object({
...
+  id: z.number().int().gte(0),
+  method: z.string(),
+  url: z.string(),
+});
+
+export type RequestSummary = z.infer<typeof RequestSummarySchema>;
 
-export type RequestorError = z.infer<typeof RequestorErrorSchema>;
+export const RequestWithResponseSchema = z.object({
+  request: z.any(),
```
stephlow, line 85: Seems like `json-schema-to-zod` has some issues with schemas referencing other schemas, sometimes it works... and in this case it doesn't. Looking into this [in another PR](https://github.com/fiberplane/fpx/pull/90) :)

stephlow:
This has been moved to #94 as somehow the git history got messed up and the PR diff is completely borked

PR #90, opened 2024-07-24
Migrate to json schemas

stephlow:
This PR changes the behavior of `cargo xtask generate-schemas` to generate JSON Schema's and their inferred types instead of Zod schemas, as unfortunately, the zod schema's didn't always work with schemas referencing other schemas.

The generated schemas also export a function that uses ajv to validate an input against the schema


fpx/src/api/models.rs
```diff
@@ -16,7 +16,7 @@ pub const FPX_WEBSOCKET_ID_HEADER: &str = "fpx-websocket-id";
 
 /// Messages that are send from the server to the client.
 #[derive(JsonSchema, Clone, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
```
hatchan, line 19: Why have you added the `deny_unknown_fields` option? This might make it impossible to have any backwards compatible models (ie. adding a new optional field will be rejected by older clients, since this new field is "unknown")

stephlow: Unfortunately this is the only way to add `additionalProperties: false` to the JSON schema's. Without this the TS types generated are not really strict, and allow for any property: e.g:

```ts
type MyType = {
   [x: string]: any,
   actualPropertyThatIsInTheSchema: bool, // etc
};
```

So the validation function using ajv still checks for it properly, but any types we add to our own functions become pretty much useless...

PR #99, opened 2024-07-29
Support form bodies in requestor

brettimus:
**THIS PR NEEDS IMMEDIATE FOLLOW UP ON THE OTEL MIDDLEWARE**

***

Adds support for requests with content types:

- `application/x-www-form-urlencoded`
- `multipart/form-data`
- `application/octet-stream`

## Issues and dark corners
- AI is not aware of how to fill in form data
- File upload UI looks a lil janky
- The non-otel middleware does not handle requests with binary data gracefully
- Replaying a request from history that had form data does not populate the body panel correctly
- Form data with a file attached will automatically become `multipart/form-data` but that's not communicated in the UI
- Show an error when we cannot proxy the request, or fail silently when some of our magic craps out (E.g., `multipart/form-data` is unparseable on our end)


frontend/src/pages/RequestorPage/FormDataForm/data.ts
```diff
@@ -0,0 +1,226 @@
...
+      return otherParameter;
+    });
+    onChange(newQueryParams);
+  };
+}
+
+/**
+ * Quick and dirty uuid utility
+ */
+function generateUUID() {
```
laulauland, line 137: nit: `crypto.randomUUID()` 🙂

laulauland (approved):
smoke tested that nothing breaks, we'll get back to the UX of this once the OTEL things are merged - LGTM (one incredibly tiny nit)

frontend/src/pages/RequestorPage/FormDataForm/data.ts
laulauland: i noticed this already exists in the current state so I'll just update it in a separate branch

PR #102, opened 2024-07-30
Add support for fpx otel traces

flenter:
- Add new `/v1/traces` URL to typescript API
    - Modify CLI to use this
    - Modify Middleware to send JSON payloads for now
- Update AI to handle different types of request bodies
- Update Requestor to work with Otel data
    - API
        - Make sure the api sends a valid otel trace id to inject as parent
    - Middleware
        - HACK - Propagate x-fpx-trace-id with a dummy span parent
- Implement route probing on middleware
    - Changed
        - Uses separate api route to report routes
        - Middleware must construct this URL itself
        - Routes are no longer refreshed with each request
- Use Hacked versions of Hono types in the middleware to prevent downstream type errors when publishing the otel middleware package
- Middleware: Report the request body and env vars in middleware root span for request
- Middleware Accept non-string messages as log arguments
- Middleware: Support patching console.debug
- Remove lots of code
    - Remove old request details page and turn on the tracing UI by default
    - Remove almost all old schemas and data types related to MizuLogs and MizuMessages, et al
    - Remove API route for `/v0/logs`


api/src/routes/app-routes.ts
```diff
@@ -62,6 +67,73 @@ app.post(
...
+              }),
+            );
+          }
+        }
+      }
+
+      return ctx.text("OK");
+    } catch (err) {
+      if (err instanceof Error) {
+        console.error("Error processing probed routes", err);
```
laulauland: nit: use `logger.error` here

api/src/routes/app-routes.ts
```diff
@@ -125,8 +197,22 @@ function serializeFormDataValue(
  * Or maybe even streams eventually?
  */
 app.all("/v0/proxy-request/*", async (ctx) => {
-  const traceId = ctx.req.header("x-fpx-trace-id") || generateUUID();
-  logger.debug("Proxying request with traceId:", traceId);
+  // Try to extract the trace id from the header, otherwise generate a new one
+  const traceIdHeader = ctx.req.header("x-fpx-trace-id");
+  const shouldUseHeaderTraceId = isValidOtelTraceId(traceIdHeader ?? "");
```
laulauland: suggestion: use `zod` and `zValidator` to validate the `proxy-request` payload, including the regex for OTEL trace id: https://zod.dev/?id=strings (`zod` supports `z.string().regex(<whatever we need here>)`)

api/src/routes/traces.ts
```diff
@@ -0,0 +1,336 @@
...
+    return ctx.text("OK");
+  } catch (error) {
+    logger.error("Error parsing trace data", error);
+    return ctx.text("Error parsing trace data", 400);
+  }
+});
+
+export default app;
+
+type MizuSpan = {
```
laulauland: maybe a nit: could we extract the below OTEL types and code into `src/lib/otel/index.ts` and `src/lib/otel/utils.ts` etc.? This will make the routes file a bit more readable and the utilities reusable.

frontend/src/hooks/useWebSocket.ts
laulauland, line 1: note: we're using `react-use-websocket` in #111

frontend/src/pages/RequestDetailsPage/RequestDetailsPageV2/Otel.tsx
```diff
@@ -0,0 +1,116 @@
...
+      orphanLogs={orphanLogs}
+      pagination={pagination}
+    />
+  );
+}
+
+/**
+ * Converts an Otel event to a so-called Orphan Log to maintain backwards compatibility with the old Mizu data format
+ */
+function convertEventToOrphanLog(
```
laulauland, line 101: any particular reason we need the backwards compatibility here?

frontend/src/pages/RequestDetailsPage/RequestDetailsPageV2/RequestDetailsPageV2Content.tsx
laulauland, line 1: small overall naming nit: any reason to keep `V2`? Since this PR moves everthing to OTEL I feel like we could simplify by just reverting to the simple `RequestDetailsPage` for all :)

frontend/src/pages/RequestDetailsPage/hooks/useNavigateToList.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    navigate("/requests");
+  });
+
+  useEffect(() => {
+    document.addEventListener("focus", handleFocus, true);
+    document.addEventListener("blur", handleBlur, true);
+    return () => {
+      document.removeEventListener("focus", handleFocus, true);
```
laulauland, line 41: more of a note to self for later: we can use `AbortController` to remove both event listeners a bit more cleanly - https://frontendmasters.com/blog/patterns-for-memory-efficient-dom-manipulation/#use-abortcontroller-to-unbind-groups-of-events

frontend/src/pages/RequestDetailsPage/v2/BodyViewerV2.tsx
```diff
@@ -0,0 +1,38 @@
...
+function FormDataViewer({ body }: { body: string }) {
+  const formData = useMemo(() => {
+    try {
+      const parsedBody = JSON.parse(body);
+      if (parsedBody && typeof parsedBody === "object") {
+        const formDataString = Object.entries(parsedBody)
+          .map(([key, value]) => {
+            if (typeof value === "object" && value !== null) {
+              // Handle nested objects or arrays
+              return `${key}=${JSON.stringify(value)}`;
```
laulauland, line 26: probably an edge case, but this will break on a circular JSON object

laulauland (approved):
a few small questions and nits but LGTM otherwise, excited to use proper trace data in our UI 🥇 🦭

frontend/src/hooks/useWebSocket.ts
brettimus: this was a hook jacco added for the rust websocket, as discussed, we can possibly remove

frontend/src/pages/RequestDetailsPage/RequestDetailsPageV2/Otel.tsx
brettimus: yeah just for the UI — it expects a flattened list of logs to render the waterfall. when the waterfall ui is more sophisticated we can remove this bit

frontend/src/pages/RequestDetailsPage/RequestDetailsPageV2/RequestDetailsPageV2Content.tsx
brettimus: no we can change the names, wanted to do that but was afraid of affecting your changes, we can do in a follow up

frontend/src/pages/RequestDetailsPage/hooks/useNavigateToList.ts
brettimus: added as comment in code

frontend/src/pages/RequestDetailsPage/v2/BodyViewerV2.tsx
brettimus: ah yes but this is dealing with something that's already been json serialized, but i will put a note

api/src/routes/traces.ts
brettimus: done and done

api/src/routes/app-routes.ts
brettimus: yeah but i guess i don't want to reject the request if the trace id is invalid, and i don't see a way to just "not allow" a header value?  could write custom middleware for this pretty easily but meh. keeping the logic in the (yes, massive) handler makes it clear where that value is coming from on first read

brettimus: well i guess you can implement fallback functionality like this https://github.com/colinhacks/zod/issues/316

api/src/routes/app-routes.ts
brettimus: jaaaa

api/src/routes/app-routes.ts
brettimus: added zvalidator and some tests for the schema to make sure it uses fallbacks correctly

PR #104, opened 2024-08-02
Add local inference server

actualwitch:
this pr adds a simple local inference server that can be used for request autofill


brettimus:
just renamed the folder from `inference` to `local-inference` so it's a little clearer from the root of the repo what the folder pertains to!

brettimus (approved):

PR #105, opened 2024-08-02
FPX Worker Socket implementation

stephlow:


fpx-workers/src/ws/worker.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        Router::with_data(self)
+            .get_async("/connect", ws_connect)
+            .post_async("/broadcast", ws_broadcast)
+            .run(req, env)
+            .await
+    }
+}
+
+async fn ws_connect(
```
hatchan, line 44: Maybe we can have a bit of documentation here describing what the purpose is, when it is going to be called, and by who (same applies to the other handler).

fpx-lib/src/api.rs
```diff
@@ -26,7 +27,7 @@ pub struct Config {
 
 #[derive(Clone)]
 pub struct ApiState {
-    _events: ServerEvents,
+    _events: Arc<dyn ServerEvents>,
```
hatchan: Think it would be nice to have `type` for this, similar to `BoxedStore` (I know the name doesn't match its implementation)

fpx-workers/src/ws/client.rs
```diff
@@ -0,0 +1,50 @@
...
+use wasm_bindgen_futures::wasm_bindgen::JsValue;
+use worker::*;
+
+pub struct WebSocketWorkerClient {
+    stub: Stub,
+}
+
+impl WebSocketWorkerClient {
+    pub fn new(env: &Arc<Env>) -> Self {
+        let stub = get_ws_durable_object(env).unwrap();
```
hatchan: We should add some error handling, instead of using unwrap/expect (unless it never fails obviously)

fpx-workers/src/ws/client.rs
```diff
@@ -0,0 +1,50 @@
...
+
+        let req = Request::new_with_init(
+            "http://fake-host/broadcast",
+            &RequestInit {
+                body: Some(payload),
+                headers: Headers::new(),
+                cf: CfProperties::new(),
+                method: Method::Post,
+                redirect: RequestRedirect::Manual,
+            },
```
hatchan: We can probably simplify this with the following: 

```suggestion
            RequestInit::new()
                .with_body(Some(payload))
                .with_method(Method::Post)
                .with_redirect(RequestRedirect::Manual),
```

fpx-workers/src/ws/client.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub fn new(env: &Arc<Env>) -> Self {
+        let stub = get_ws_durable_object(env).unwrap();
+
+        Self { stub }
+    }
+
+    pub async fn connect(&self) -> Result<Response> {
+        let mut req = Request::new("http://fake-host/connect", Method::Get)?;
+
+        req.headers_mut()?.set("Upgrade", "websocket")?;
```
hatchan:
```suggestion
        let headers = Headers::from_iter([("Upgrade", "websocket")]);
        let req = Request::new_with_init(
            "http://fake-host/connect",
            RequestInit::new().with_headers(headers),
        )?;
```

fpx-workers/src/ws/handlers.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+#[worker::send]
+pub async fn ws_connect(State(state): State<WorkerApiState>, headers: HeaderMap) -> Response<Body> {
+    if let Some(value) = headers.get("Upgrade") {
+        if value == "websocket" {
+            return WebSocketWorkerClient::new(&state.env)
+                .connect()
+                .await
+                .unwrap()
```
hatchan: There are some `unwrap`s here as well that we should handle.

fpx-workers/src/lib.rs
```diff
@@ -33,14 +41,49 @@ async fn fetch(
...
-    let events = ServerEvents::new();
 
-    let service = service::Service::new(boxed_store.clone(), events.clone());
-    let mut router = api::create_api(events, service, boxed_store);
+    let service = service::Service::new(boxed_store.clone(), boxed_events.clone());
+    let api_router = api::create_api(boxed_events, service, boxed_store);
+
+    let mut router: axum::Router = axum::Router::new()
+        .route("/api/ws", get(ws_connect))
+        .route("/api/ws/broadcast", post(ws_broadcast))
```
hatchan: This endpoint would be exposed to the public, but it should only be defined on the durable object worker, right?

fpx-workers/src/ws/worker.rs
```diff
@@ -0,0 +1,62 @@
+use fpx_lib::api::models::ServerMessage;
+use worker::*;
+
+/// Based on:
+/// https://developers.cloudflare.com/durable-objects/examples/websocket-hibernation-server/
+#[allow(dead_code)]
```
hatchan: Is this still required?

fpx-workers/src/ws/handlers.rs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    Response::builder()
+        .status(StatusCode::UPGRADE_REQUIRED)
+        .body(Body::from("Durable Object expected Upgrade: websocket"))
+        .unwrap()
+}
+
+#[worker::send]
+pub async fn ws_broadcast(
```
hatchan: Related to the comment in the router, this can then also be removed, right?

fpx-workers/src/lib.rs
stephlow: Oops, this was an old testing endpoint! Removed it :)

hatchan (approved):

PR #106, opened 2024-08-02
Fix sending of json payloads to proxy request

brettimus:
title says it all -


frontend/src/pages/RequestorPage/queries.ts
```diff
@@ -225,7 +225,7 @@ export function makeProxiedRequest({
 function createBody(body: RequestorBody) {
   if (body.type === "json") {
     if (typeof body.value !== "undefined") {
-      return JSON.stringify(body.value);
+      return body.value;
```
brettimus, line 228: this is already a string!

laulauland (approved):

PR #111, opened 2024-08-08
Add webhonc proxy service

laulauland:
TODOs:
- [x] revise the interaction flow for copying the URL
- [x] add a settings feature flag

This PR introduces the webhonc proxy feature - which allows you to forward requests made to a temporary public URL against your own API through FPX.

The PR introduces a new package called `webhonc` that is a small utility service deployed on Cloudflare's platform using Durable Objects for stateful websocket connections.

How it works:

1. User spins up `fpx` and establishes a websocket connection with `webhonc` and receives an `id` (+url) for their connection.
2. User can make requests from the public network against that URL (e.g.: webhook requests from GitHub or elsewhere).
3. The request information gets captured in the `webhonc` service and forwarded "down the wire" to the connected `fpx` instance.
4. `fpx` recreates that requests and runs it against the user's API in local dev mode.
5. `fpx` captures and records response from the API as per usual and shows the request/response as proxied in the UI
6. `fpx` does NOT proxy the response back out to the original service - the original service will always get a `200 OK` response. This was cut to reduce scope for the initial release but will be introduced in a later update.

 Things to review/monitor/consider:
- feasibility of using Durable Objects as a platform for the proxy service. Although best practices from Cloudflare regarding websocket server hibernation have been applied, we should still monitor the costs for this.
- some sort of authentication should probably be established for the final rollout of the feature, especially if the proxied response feature is rolled out.

Notable changes:
- some of the schemas for data payloads are starting to be extracted into `api/src/lib/types.ts` - this will become a separate package in the upcoming `pnpm` monorepo setup;
- initial `api` refactoring to support reusable `executeRequest` that can be triggered both from the UI and from `webhonc` .


api/src/logger.ts
```diff
@@ -2,10 +2,14 @@
...
+
 // Determine the current log level from the environment variable or default to "info"
-const currentLogLevel: LogLevel = isLogLevel(process.env.FPX_LOG_LEVEL)
-  ? process.env.FPX_LOG_LEVEL
-  : "info";
+const currentLogLevel: LogLevel = isDev
+  ? "debug"
+  : isLogLevel(process.env.FPX_LOG_LEVEL)
+    ? process.env.FPX_LOG_LEVEL
+    : "info";
```
brettimus: Could still be nice to turn off debug logging in dev, so we can simulate the end user terminal output

```suggestion
const defaultLogLevel: LogLevel = isDev ? "debug" : "info";
const currentLogLevel: LogLevel = isLogLevel(process.env.FPX_LOG_LEVEL)
    ? process.env.FPX_LOG_LEVEL
    : defaultLogLevel;
```

frontend/src/components/WebhoncBadge/WebhoncBadge.tsx
```diff
@@ -0,0 +1,81 @@
...
+        align="end"
+        sideOffset={5}
+        className="rounded-xl border bg-background z-10"
+      >
+        <div className="rounded-xl bg-muted/40">
+          <div className="grid gap-4 p-4">
+            <div className="font-normal text-sm w-[400px] flex flex-col gap-2">
+              <p className="text-wrap">
+                <span className="font-semibold">Proxy URL:</span> any request
+                received at this URL will be forwarded to the FPX and your API.
```
brettimus:
```suggestion
                received at this URL will be forwarded to your API.
```

brettimus: copy tweak - probably don't have to mention it's forwarded to fpx? but if we want to, we could say "will be forwarded via fpx to your api"

api/src/lib/webhonc/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  const protocol = host.startsWith("localhost") ? "ws" : "wss";
+  const socket = new WebSocket(`${protocol}://${host}/ws`);
+
+  socket.onopen = () => {
+    logger.debug(`Connected to the webhonc service at ${host}`);
+  };
+
+  socket.onmessage = async (event) => {
+    logger.debug("Received message from the webhonc service:", event.data);
+    await handleMessage(event, wsConnections, db);
```
brettimus:
```suggestion
    try {
      await handleMessage(event, wsConnections, db);
    } catch (err) {
      logger.debug("Webhonc handleMessage had an error", err);
    }
```

this can throw, might be nice to print the error. could also see if it's an instance of a zod error (not sure what the thrown error looks like?) and format the error string accordingly

api/src/lib/webhonc/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  };
+
+  socket.onclose = (event) => {
```
brettimus, line 42: for the future: could be good to maintain some state about the webhonc connection and propagate that to the UI somehow

api/src/lib/webhonc/index.ts
```diff
@@ -0,0 +1,161 @@
...
+
+const messageHandlers: {
+  [K in z.infer<typeof WsMessageSchema>["event"]]: (
+    message: Extract<z.infer<typeof WsMessageSchema>, { event: K }>,
+    wsConnections: Set<WebSocket>,
+    db: LibSQLDatabase<typeof schema>,
+  ) => Promise<void>;
+} = {
+  trace_created: async (_message, _wsConnections, _db) => {
+    logger.info("trace_created message received, no action required");
```
brettimus:
```suggestion
    logger.debug("trace_created message received, no action required");
```

probably don't want this to show up in the end user's terminal

api/src/lib/webhonc/index.ts
```diff
@@ -0,0 +1,161 @@
...
+        JSON.stringify({
+          event: "connection_open",
+          payload: { connectionId },
+        }),
+      );
+    }
+  },
+  request_incoming: async (message, _wsConnections, db) => {
+    // no trace id is coming from the websocket, so we generate one
+    const traceId = crypto.randomUUID();
```
brettimus, line 107: todo: change to an otel id after merge

api/src/lib/webhonc/index.ts
```diff
@@ -0,0 +1,161 @@
...
+        requestHeaders: message.payload.headers,
+        // @ts-expect-error - Trust me, the request method is correct, and it's a string
+        requestMethod: message.payload.method,
+        requestBody: message.payload.body,
+        requestUrl,
+      });
+
+      const duration = Date.now() - startTime;
+
+      await handleSuccessfulRequest(db, requestId, duration, response);
```
brettimus, line 161: note - this will need to accept the traceId from above, since the fpx otel library no longer returns the traceId in its headers ((i tried to do that and failed, it's harder now that we are no longer middleware))

brettimus: that change is already on my otel branch though, so don't worry about making it here

api/src/index.node.ts
```diff
@@ -67,26 +81,13 @@ const watchDir = process.env.FPX_WATCH_DIR ?? process.cwd();
...
-  ws.on("close", (code) => {
-    wsConnections.delete(ws);
-    logger.debug("WebSocket connection closed", code);
-  });
-});
+// check settings if proxy requests is enabled
+const proxyRequestsEnabled = await getSetting(db, "proxyRequestsEnabled");
+
+if (proxyRequestsEnabled) {
+  logger.info("Proxy requests feature enabled, connecting to webhonc...");
```
brettimus:
```suggestion
  logger.debug("Proxy requests feature enabled, connecting to webhonc...");
```

`info` will log to the user's terminal fwiw, probably want debug here?

frontend/src/hooks/useProxyRequestsEnabled.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Hook that reads the user's settings to determine if they proxy requests enabled
+ * Defaults to false
+ *
+ */
+export function useProxyRequestsEnabled() {
+  const { data } = useFetchSettings();
+  return useMemo(() => {
+    console.log(data);
```
brettimus, line 12: cl

frontend/src/hooks/useWebsocketQueryInvalidation.ts
```diff
@@ -1,63 +1,51 @@
...
   const queryClient = useQueryClient();
 
-  useEffect(() => {
-    const socket = new WebSocket("/ws");
+  const wsMessage = useRealtimeService();
+
+  if (wsMessage) {
+    switch (wsMessage.event) {
+      case "trace_created": {
+        console.debug("trace_created");
```
brettimus, line 19: cl

webhonc/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "name": "@fiberplane/webhonc",
+  "description": "Webhonc is a simple service running on Durable Objects",
+  "scripts": {
+    "dev": "wrangler dev",
+    "deploy": "wrangler deploy --minify",
+    "lint:ci": "biome ci .",
+    "cf-typegen": "wrangler types --env-interface CloudflareBindings"
+  },
+  "dependencies": {
+    "@fiberplane/hono": "^0.2.0-beta.1",
```
brettimus: rm?

frontend/src/components/WebhoncBadge/WebhoncBadge.tsx
```diff
@@ -0,0 +1,84 @@
+import { cn } from "@/utils";
+// import { Popover, PopoverContent, PopoverTrigger } from "../ui/popover";
```
brettimus, line 2: rm

brettimus:
```suggestion
```

rm?

frontend/src/components/WebhoncBadge/WebhoncBadge.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { useWebhoncConnectionId } from "./queries";
+
+export function WebhoncBadge() {
+  const [isCopied, setIsCopied] = useState(false);
+
+  const { data: url, isPending } = useWebhoncConnectionId();
+
+  const handleCopy = () => {
+    // copy to clipboard
+    navigator.clipboard.writeText(url!);
```
brettimus, line 20: nit: this doesn't handle the error case where text wasn't copied. we can refactor later on my branch since i have a utility hook for copying to clipboard

frontend/src/hooks/useProxyRequestsEnabled.ts
brettimus:
```suggestion
```

rm?

webhonc/src/index.ts
```diff
@@ -0,0 +1,94 @@
...
+
+app.all(
+  "/:id/*",
+  zValidator("param", z.object({ id: z.string() })),
+  async (c) => {
+    const { id } = c.req.valid("param");
+
+    // We need to serialize the path components without the id
+    // so we can send it down the wire and replay it locally
+    const pathComponentsWithoutId = c.req.path
```
brettimus, line 42: still might prefer something that just replaces `/^\/id/`, feels more straightforward

webhonc/src/index.ts
```diff
@@ -0,0 +1,94 @@
...
+      body = null;
+    } else {
+      switch (contentType) {
+        case "application/json":
+          body = await c.req.json();
+          break;
+        case "application/x-www-form-urlencoded":
+          body = await c.req.formData();
+          break;
+        case "text/plain":
```
brettimus: i know it's not as elegant, but we might wanna prefer an if-else chain with `contentType?.includes(...)` instead of a switch statement.

brettimus: reason being that sometimes people will add additional info to the content type like `text/html; charset=UTF-8`

webhonc/src/index.ts
```diff
@@ -0,0 +1,94 @@
...
+          body = await c.req.text();
+          break;
+        default:
```
brettimus: additional cases i think we should cover:

- `text/html`
- `application/xml`

webhonc/src/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    const doId = c.env.WEBHONC.idFromString(id);
+    const webhonc = c.env.WEBHONC.get(doId) as DurableObjectStub<WebHonc>;
+
+    const headers = c.req.raw.headers;
+    const headersJson: { [key: string]: string } = {};
+    for (const [key, value] of headers.entries()) {
+      // these are calculated dynamically by the fetch client so we don't send them down the wire
+      if (
+        key.toLowerCase() === "content-length" ||
+        key.toLowerCase() === "content-type"
```
brettimus: wait i don't think you want to abandon the `content-type` here? unless i'm missing something. 

when i send a request via the webhonc with `application/json` as the content-type, it gets converted to a `text/plain;charset=utf-8` header by the fetch call in fpx

brettimus: to be clear, you **SHOULD** abandon the content-type header for multipart form data, since the form data boundary is generated dynamically by fetch afaik, but we're not handling multipart requests here anyhow

webhonc/src/index.ts
```diff
@@ -0,0 +1,94 @@
...
+import { WebHonc } from "./webhonc";
+
+const app = new Hono<{ Bindings: Bindings; Variables: Variables }>();
+
+app.get("/ws", async (c) => {
+  if (c.req.header("upgrade") !== "websocket") {
+    return new Response("Not a websocket request", { status: 426 });
+  }
+
+  const id = c.env.WEBHONC.newUniqueId();
```
brettimus, line 18: question: what happens to old ids? are they still persisted somewhere? should we dispose of them when the websocket connection closes?

webhonc/src/index.ts
```diff
@@ -0,0 +1,94 @@
...
+  const webhonc = c.env.WEBHONC.get(id) as DurableObjectStub<WebHonc>;
```
brettimus, line 19: TODO - handle the case where nothing is returned? right now if i pass in a random id to this route, we just get an internal server error. a 404 could be nicer

webhonc/src/index.ts
brettimus: html is a little ridiculous but ya know. doesn't hurt to support it. just need to parse the body as text for both of those cases

webhonc/src/index.ts
laulauland: they get disposed in Cloudflare's runtime I believe. At least there's no documentation on clean up work needed AFAIK

api/src/index.node.ts
laulauland: to some extent we'd want to inform the user that `fpx` is making an outside connection but this is kind of evident in the UI so changing it to `debug`

api/src/lib/webhonc/index.ts
laulauland: yea that's a good point, adding a `TODO:` comment for now

webhonc/src/index.ts
laulauland: tbh I experimented a little bit here and found even the simplest regex approach a bit of a pain to read. I agree that sending path components as array items might be overkill so I can just join them back into a string before sending down?

webhonc/src/index.ts
laulauland: yea that's a good call, refactored

webhonc/src/index.ts
laulauland: yea not sure why I decided to wipe `Content-Type` - changing this

webhonc/src/index.ts
laulauland: here it shouldn't be a problem because id is created using DO's `newUniqueId()`, however, in the other route it is indeed a problem

webhonc/src/index.ts
laulauland: handled (even csv's now)

webhonc/src/index.ts
brettimus: joining them could be good but let's do that later, eager to merge this

webhonc/src/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+      contentType === "text/css" ||
+      contentType === "text/csv"
+    ) {
+      console.debug(
+        "Content type is XML, HTML, JavaScript, CSS, or CSV, returning text",
+      );
+      return await c.req.text();
+    }
+
+    // TODO: Handle binary data (application/octet-stream)
```
brettimus, line 74: okay let's do this in a follow-up, but when we **know** we're dealing with binary data, including audio/video/image/pdf, we probably don't want this default case of `.text()`, since it will return a big ole text rendition of the binary data

same goes for multipart form data

for the use case we're targeting, JSON webhook payloads, we'll be fine, however 👍

brettimus (approved):

webhonc/src/utils.ts
laulauland: yea agree, default text serialization is definitely not optimal

PR #112, opened 2024-08-08
Refactor fpx to use the new fpx_lib

hatchan:
Create implementation for Store and ServerEvents, both use their
original implementation with some caveats.

Delete all unused components. These components might be returned in
a future PR, but it is easier to now get rid of it, instead of having to
worry about backwards compatibility.


xtask/src/commands/schemas.rs
```diff
@@ -20,14 +19,8 @@ pub async fn handle_command(args: Args) -> Result<()> {
...
         schema_for!(ClientMessage),
-        schema_for!(NewRequest),
-        schema_for!(NewRequestError),
-        schema_for!(Request),
-        schema_for!(RequestAdded),
-        schema_for!(RequestSummary),
-        schema_for!(RequestWithResponse),
-        schema_for!(Response),
         schema_for!(ServerMessage),
+        // schema_for!(fpx_lib::api::models::Span),
```
stephlow, line 23: Probably a good idea to re-run the task so the old schemas and types are removed from TypeScript as well

stephlow (changes requested):

xtask/src/commands/schemas.rs
hatchan: Applied in `bd5ee9467bfe5ee7e2733cbfb48fe8bff8dc0261`

stephlow (approved):
🙌

PR #113, opened 2024-08-08
Allow fetching the summary of a single trace

hatchan:


fpx/src/api/client.rs
```diff
@@ -55,7 +58,7 @@ impl ApiClient {
     ) -> Result<T, ApiClientError<E>>
     where
         T: serde::de::DeserializeOwned,
-        E: serde::de::DeserializeOwned,
+        E: serde::de::DeserializeOwned + Error,
```
hatchan, line 61: This is now required since we want to set the `status_message` in the Otel trace from the actual Error.

stephlow (approved):

PR #117, opened 2024-08-09
Fix serialization of log messages (and args) that are attached to otel event attrs

brettimus:
The issue had to do with the fact that `Error` does not JSON serialize. Just shows up as an empty object.


laulauland (approved):
wheee

PR #120, opened 2024-08-12
Improve headers tables

brettimus:
This PR will include:

- Collapsible headers tables, so we don't render like 20 rows when you don't need to see all of them immediately
- Blurred out sensitive headers with a UI to show them

### DIFF
--- a/frontend/src/pages/RequestDetailsPage/v2/KeyValueTableV2.tsx
+++ b/frontend/src/pages/RequestDetailsPage/v2/KeyValueTableV2.tsx
@@ -2,43 +2,25 @@ import { Table, TableBody, TableCell, TableRow } from "@/components/ui/table";
 import { cn } from "@/utils";
 import { useCallback, useMemo, useState } from "react";
 import { Collapsible, CollapsibleTrigger, CollapsibleContent } from "@/components/ui/collapsible";
-import { Button } from "@/components/ui/button";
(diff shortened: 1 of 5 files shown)


mies (approved):

frontend/src/pages/RequestDetailsPage/v2/KeyValueTableV2.tsx
```diff
@@ -39,3 +104,51 @@ export function KeyValueTableV2({
...
+  keyValue: Record<string, string>;
+  emptyMessage?: string;
+  className?: string;
+  defaultCollapsed?: boolean;
+  title: string;
+  sensitiveKeys?: string[];
+}) {
+  const [isOpen, setIsOpen] = useState(!defaultCollapsed);
+  const count = useMemo(() => Object.entries(keyValue).length, [keyValue]);
+  const toggleIsOpen = useCallback(() => setIsOpen((o) => !o), []);
```
oscarvz: Why is this memoized? It's probably introducing more overhead than performance gains for these light functions/expressions.

oscarvz (approved):
One nit, lgtm!

frontend/src/pages/RequestDetailsPage/v2/KeyValueTableV2.tsx
brettimus: ah right, will unmemo

PR #122, opened 2024-08-12
Modify FPX cli to add nodejs compatibility flags to cloudflare projects

brettimus:
FP-3939


brettimus:
tested with:

- Cf project with existing compat flags, but no nodejs compat flag
- Cf project without compat flags 
- Cf project with nodejs compat flag already defined
- Deno project (which skips the new step in CLI)

mies (approved):

PR #126, opened 2024-08-13
Use workspace package for shared types

stephlow:


pnpm-workspace.yaml
laulauland, line 1: could we move the types package inside packages/*?

laulauland (approved):
small thing but otherwise lgtm

PR #127, opened 2024-08-13
Show responses from proxied requests in requestor, even if request is not to linked service

brettimus:
Summary:

- Render response AS-IS (when we can) for requests made through requestor
  - For binary responses, we only handle images right now
  
- Allow requests to services other than the target service for FPX
  - This required A TON of changes since it broke a fundamental assumption of the UI, that you're only testing your service

- Simplify `RequestorInput` to be fully controlled by the `path` state inside the main Requestor reducer
  - This required reconfiguring how we automagically add the service baseUrl to requests, as well as how we match history entries to the form input fields

- Remove the `Headers` tab from the response panel and put headers as a collapsed bit above the response body

***

To test:

- Go through your usual flow, make sure nothing broke
- Make requests to a non-monitored service, and make sure the responses render as expected
- Try loading entries from history
- Try filling in request parameters, generating a prompt for testing, etc

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.7.0-unstable.2",
+  "version": "0.7.0-alpha.6",
   "name": "@fiberplane/studio",
   "description": "Local development debugging interface for Hono apps",
   "author": "Fiberplane<info@fiberplane.com>",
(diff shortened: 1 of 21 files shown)


laulauland:
we can resolve this in a separate PR but just noting that in the history list there's no distinction between requests fired off against the monitored API and the guest APIs
CleanShot 2024-08-15 at 11 43 22@2x

laulauland (approved):

brettimus:
> we can resolve this in a separate PR but just noting that in the history list there's no distinction between requests fired off against the monitored API and the guest APIs CleanShot 2024-08-15 at 11 43 22@2x

added linear tix

PR #129, opened 2024-08-13
Improvements on the URL proxying

laulauland:
This is a set of improvements to the URL proxying feature (codenamed "webhonc").

- [x] store the URL proxy ID in the database and reuse it when the same URL is requested again. This is to avoid having to update the proxy URL every time you startup `@fiberplane/studio`
- [x] UI updates to make it easier to understand how the proxy URL works
- [x] allow `@fiberplane/studio` to live reconnect to the proxy service when the setting is toggled

A couple of other improvements:
 - chore: `drizzle-*` things are updated
 - settings are stored as key-value pairs in the db instead of a single JSON string


api/drizzle/0009_ambiguous_ogun.sql
```diff
@@ -0,0 +1,5 @@
+DELETE FROM `settings`;--> statement-breakpoint
```
brettimus, line 1: needed to add this otherwise migration will fail if you have an existing settings record.

api/src/index.node.ts
```diff
@@ -83,11 +82,16 @@ startRouteProbeWatcher(watchDir);
...
 
+// set up webhonc manager
+webhonc.setupWebhonc({ host: resolveWebhoncUrl(), db, wsConnections });
+
 // check settings if proxy requests is enabled
-const proxyRequestsEnabled = await getSetting(db, "proxyRequestsEnabled");
+const proxyRequestsEnabled = (await getSetting(
+  db,
+  "proxyRequestsEnabled",
+)) as boolean;
```
brettimus: can avoid the ts coercion and just make this value a bool with `!!`

```suggestion
const proxyRequestsEnabled = !!(await getSetting(
  db,
  "proxyRequestsEnabled",
));
```

api/src/lib/settings/index.ts
```diff
@@ -0,0 +1,182 @@
...
+      return acc;
+    },
+    {},
+  );
+}
+
+// HACK: This is a temporary solution to handle boolean and number values,
+// since all the settings are currently stored as strings.
+// Revisit this once db schema is updated
+function normalize(value: string): boolean | number | string {
```
brettimus, line 63: this is a little smelly and was why i shied away from using denormalized columns. i don't *think* we will hit any issues with current settings, but we need to be careful if we store things like the ai temperature in settings table, as a value of `1` or `1.0` would get converted to `true`

api/src/lib/webhonc/index.ts
```diff
@@ -15,26 +15,87 @@ import {
...
+  await connect();
+}
+
+export async function stop() {
+  if (socket && socket.readyState === WebSocket.OPEN) {
+    socket.close(1000, "Closing connection due to settings change");
+    socket = undefined;
+  }
+  if (reconnectTimeout) {
+    console.log("closing reconnect timeout");
```
brettimus: console log — use logger instead

api/src/lib/webhonc/index.ts
```diff
@@ -15,26 +15,87 @@ import {
...
+    logger.debug("Webhonc process already started. Skipping...");
+    return;
+  }
+
+  logger.debug("Starting Webhonc process...");
+  await connect();
+}
+
+export async function stop() {
+  if (socket && socket.readyState === WebSocket.OPEN) {
```
brettimus: edge case - what happens if you try to `stop` when the websocket is connecting or closing? 

actual question, not rhetorical. i'm not clear if it will emit an error if you try to close a non-open state

if it won't throw errors in that case, we could just call `close` if the socket exists

api/src/lib/webhonc/store.ts
```diff
@@ -1,12 +1,18 @@
...
-// whenever the app boots up and establishes a connection with the webhonc
-// service.
-let webhoncConnectionId: string | undefined = undefined;
+import type { LibSQLDatabase } from "drizzle-orm/libsql";
+import type * as schema from "../../db/schema.js";
+import { getSetting, upsertSettings } from "../settings/index.js";
 
-export function setWebHoncConnectionId(id: string) {
-  webhoncConnectionId = id;
+const WEBHONC_CONNECTION_ID_KEY = "webhoncConnectionId";
```
brettimus, line 5: should this constant live in settings to keep all settings together?

api/src/lib/webhonc/index.ts
brettimus: you don't have to do anything about this comment btw, i don't think we have to deal with this edgy edge case rn, just came to mind while reviewing

laulauland: indeed verified that simply calling `socket.close()` if it exists should be fine, as the `.close()` will be a no-op if the state is already closed https://github.com/websockets/ws/blob/019f28ff1ffddfcdc428d1de5ecd98648057a2ab/lib/websocket.js#L259

api/src/lib/webhonc/store.ts
laulauland: I kind of keep it localized because really only the `get...()` and `set...()` functions are used to access it but I can revisit that

brettimus (approved):

mies (approved):

PR #137, opened 2024-08-15
Use string attributes in otel middleware

actualwitch:


laulauland:
Looks good! Could you have a look at updating the [OTEL helpers](https://github.com/fiberplane/fpx/blob/main/frontend/src/pages/RequestDetailsPage/v2/otel-helpers.ts) in the UI? I think they currently expect a number so will need to accommodate for it.

actualwitch:
> Looks good! Could you have a look at updating the [OTEL helpers](https://github.com/fiberplane/fpx/blob/main/frontend/src/pages/RequestDetailsPage/v2/otel-helpers.ts) in the UI? I think they currently expect a number so will need to accommodate for it.

added!

frontend/src/pages/RequestDetailsPage/v2/otel-helpers.ts
```diff
@@ -90,6 +90,10 @@ export function getNumber<T = number>(
     defaultValue: T;
   },
 ) {
+  if (typeof value === "string") {
```
brettimus: nit: Lau and discovered today that `Number("")` returns 0. i'm going to add a case that this is not an emptystring. then, approved!

brettimus (approved):

PR #143, opened 2024-08-17
First pass: Collapsible text view

brettimus:
## Changes

For fetch and request spans, if the body is a loonnngg block of text, truncate it and allow the user to expand.

This functionality must be explicitly enabled on the text viewer component, so unless otherwise specified, behavior has not changed for existing text viewer component in the UI. 

## Examples

Long text response body

image

### DIFF
--- a/frontend/src/pages/RequestorPage/ResponsePanel.tsx
+++ b/frontend/src/pages/RequestorPage/ResponsePanel.tsx
@@ -556,14 +556,16 @@ export function ResponseBodyText({
   defaultExpanded?: boolean;
   className?: string;
 }) {
+  const [isExpanded, setIsExpanded] = useState(!!defaultExpanded);
+  const toggleIsExpanded = () => setIsExpanded((e) => !e);
+
   // For text responses, just split into lines and render with rudimentary line numbers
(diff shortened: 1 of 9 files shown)


mies (approved):

frontend/src/pages/RequestorPage/ResponsePanel.tsx
```diff
@@ -545,31 +545,118 @@ function ResponseBodyBinary({
...
+          >
+            {isExpanded ? "Show Less" : "Show More"}
+          </Button>
+        </div>
+      )}
     </div>
   );
 }
 
+function useTextPreview(
```
flenter, line 603: Doesn't need to be part of this pr, but this file is really ripe to be split up into separate parts: just make a directory

The ResponseBodyText component for instance could benefit from its own file especially since it's been exported.

flenter (approved):

frontend/src/pages/RequestorPage/ResponsePanel.tsx
brettimus: did it!

PR #149, opened 2024-08-19
Add .js to all relevant otel middleware imports

brettimus:
we should just get rid of swc


actualwitch (approved):

flenter:
I've opened up a PR https://github.com/fiberplane/fpx/pull/150 that also solves the imports without requiring us to use the ".js" extension for imports in the typescript code

brettimus:
closing in favor of https://github.com/fiberplane/fpx/pull/150/files

PR #150, opened 2024-08-19
Add tsc-alias

flenter:
This updates the imports in the compiled code to include extensions

So for instance `dist/index.js` for the otel library changes from:

``` js
export { instrument } from "./instrumentation";
export { measure } from "./measure";

//# sourceMappingURL=index.js.map% 
```
to:

``` js
export { instrument } from "./instrumentation.js";
export { measure } from "./measure.js";

//# sourceMappingURL=index.js.map%  
```


brettimus (approved):
much better than my pr. good to have you back.

PR #152, opened 2024-08-19
Fix issue with node envs (must explicitly set `duplex` option on Request constructror)

brettimus:
Alright. So. In environments that use node/undici, we need to conform to specifications APPARENTLY. 

- https://github.com/nodejs/node/issues/46221
- https://github.com/whatwg/fetch/pull/1457

***

The fetch specification is in a weird spot. 

Technically, when the body of a `RequestInit` is a ReadableStream, you **MUST** specify the property `duplex: "half"`

There will be future support for `duplex: "full"`, however this is not yet exposed. 

So really, the only option for `duplex` is `"half"` but we need to use it explicitly for future compatibility.


brettimus:
tested by Nele on her/oscar's bun project, where the issue first came up

frontend/src/pages/RequestorPage/ResponsePanel.tsx
```diff
@@ -426,7 +426,7 @@ function ResponseBody({
     }
 
     if (body?.type === "json") {
-      const prettyBody = JSON.stringify(JSON.parse(body.value), null, 2);
+      const prettyBody = JSON.stringify(safeParseJson(body.value), null, 2);
```
brettimus, line 429: shoehorned this in here, will be useful if we ever get a partial response back from a server for some reason ((happened to me, not sure why))

brettimus: the alternative here is that we crash the entire app

oscarvz (approved):
Can confirm it works on the bun project 👍

frontend/src/pages/RequestorPage/ResponsePanel.tsx
```diff
@@ -660,3 +660,13 @@ function FailedWebsocket() {
     </div>
   );
 }
+
+const safeParseJson = (jsonString: string) => {
```
flenter, line 664: There's already an ~eerily similar~ exact same `safeParseJson` function in `src/utils/index.ts`. Why not import & use that?

flenter (commented):
Oh it's been merged  😅  I didn't finish my review fast enough . I noticed one small thing though

PR #153, opened 2024-08-19
Add copy as cURL button to Requestor

oscarvz:
Adds a 'copy as cURL' button to the bottom bar in Requestor. It's currently behind a feature flag still.

- Add `BottomToolbar` directory that houses its child components:
  - New: `CopyAsCurl`
  - Moved: `RequestBodyCombobox`
- Logic & user feedback when a button is disabled (based on method and/or unsupported types)

> [!note]
> I didn't bother with memoizing any of the logic implemented in `CopyAsCurl`, as every single interaction will result in a rerender - unless we memoize each prop individually, which introduces quite some overhead with negligible performance gains.

---
Can do:
https://github.com/user-attachments/assets/c2fe3a08-b39e-4b02-9ee6-b1bb9d9fe4f0

No can do:
https://github.com/user-attachments/assets/a19a7b31-7c59-414d-9a8c-02ceade66429


frontend/src/pages/RequestorPage/RequestPanel/BottomToolbar/CopyAsCurl.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const isUnsupportedBodyType =
+    body.type === "file" || (body.type === "form-data" && body.isMultipart);
+
+  const handleCopy = async () => {
+    const payload = getBodyValue({ body, method });
+
+    // As we don't support automatic body type switching yet, we now manually
+    // set the initial header the correspoinding content type based on the
+    // method & body. Otherwise, it will be empty.
+    const initialHeader = payload ? getContentTypeHeader(body) : "";
```
oscarvz: Depends on
- #168

oscarvz: Putting PR back into draft as above's PR is merged, resulting in duplicate headers being added to the cURL command.

frontend/src/pages/RequestorPage/RequestPanel/BottomToolbar/CopyAsCurl.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    const url = new URL(path);
+    for (const { enabled, key, value } of queryParams) {
+      if (enabled) {
+        url.searchParams.append(key, value);
+      }
+    }
+
+    const data = payload ? `-d ${payload}` : "";
+    const curlCommand = `curl -X ${method} ${url} ${headers} ${data}`;
```
brettimus: might wanna put the url in quotes, especially if there's a querystring attached, since most shells need a `?` to be wrapped in quotes!

to test locally, try making a GET request against a route with some query params attached

frontend/src/pages/RequestorPage/RequestPanel/BottomToolbar/CopyAsCurl.tsx
```diff
@@ -0,0 +1,93 @@
...
+}: CopyAsCurlProps) {
+  const [isCopied, setIsCopied] = useState(false);
+  const timeout = useRef<ReturnType<typeof setTimeout>>();
+  const isUnsupportedBodyType =
+    body.type === "file" || (body.type === "form-data" && body.isMultipart);
+
+  const handleCopy = async () => {
+    const payload = getBodyValue({ body, method });
+
+    const headers = requestHeaders.reduce((acc, { enabled, key, value }) => {
```
brettimus, line 36: i'll file an issue for this, but one thing that struck me is that with my silly "automatic content-type header" changes, we attach a content-type to GET requests, which isn't necessary.

but yeah. that's an upstream issue, i can fix it

frontend/src/pages/RequestorPage/RequestPanel/BottomToolbar/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+  // Prevent sending JSON body for GET and HEAD requests, as they're the only
+  // methods where providing data is invalid.
+  if (method === "GET" || method === "HEAD" || !body.value) {
+    return undefined;
+  }
+
+  const bodyValue = (() => {
+    switch (body.type) {
+      case "form-data": {
+        return body.value.reduce((acc, field) => {
```
brettimus: so instead of constructing this manually, you could also make use of `URLSearchParams` and that would help you get encoding for free. i should've thought of this when we talked yesterday but it just hit me 

```ts
import { FormDataParameter } from "../../FormDataForm/types";

export function formDataToUrlEncoded(formDataParams: FormDataParameter[]) {
  const urlEncodedFormData = new URLSearchParams();

  formDataParams.forEach((field) => {
    if (field.enabled && field.value.type === "text") {
      urlEncodedFormData.append(field.key, field.value.value);
    }
  });

  return urlEncodedFormData.toString();
}
```

brettimus: then in your switch case just call `formDataToUrlEncoded(body.value)`

frontend/src/pages/RequestorPage/RequestPanel/BottomToolbar/CopyAsCurl.tsx
oscarvz: Good one! Will add

frontend/src/pages/RequestorPage/RequestPanel/BottomToolbar/utils.ts
oscarvz: This indeed is a nicer way to approach it. Will add!

frontend/src/pages/RequestorPage/RequestPanel/RequestPanel.tsx
```diff
@@ -79,7 +79,7 @@ function ResizableRequestMeta(props: RequestPanelProps) {
   const styleWidth = useStyleWidth(width);
   return (
     <Resizable
-      className="min-w-[200px] overflow-hidden h-full"
```
brettimus: without the `overflow-hidden` here we don't get scrollability for the inner contents when they overflow (many request params, long req body)

brettimus: then once we do have a scrollable container, might need to add some bottom padding: 
image

brettimus (approved):
pre-approving, but we gotta fix the issues with vertically scrolling through overflowing content in the request panel before smashing that merge button. left a comment on the relevant code

PR #155, opened 2024-08-19
Support fpx worker ingestion engine in studio

stephlow:
- Add types and endpoints to the Rust `fpx-workers` that are compatible with the TypeScript api
- Consolidate the different settings schemas for both the backend and the frontend and make sure all handlers are properly typed
- Removed `raw_payload` column from the `otel_spans` table, as it contains the trace with all the spans as [discussed here](https://fiberplane.slack.com/archives/C075HHYANQM/p1723811323065569)
- Some general improvements to types and code style
- Validate settings before publishing them to the db
- Serialize/Deserialize settings as JSON so one setting can have multiple values / types
- Moved the api key clearing from frontend to backend
- Generated Rust -> TypeScript schemas are now bundled with the `fpx-types` package instead of just the `frontend` package
- Add a proxy in the TypeScript api that forwards traces / span requests to an external fpx worker
- Add settings form for configuring FPX worker proxy


frontend/src/queries/settings.ts
```diff
@@ -1,24 +1,40 @@
...
 
 const FPX_USER_SETTINGS_QUERY_KEY = "FPX_USER_SETTINGS";
 
 export function useFetchSettings() {
   return useQuery({
     queryKey: [FPX_USER_SETTINGS_QUERY_KEY],
-    queryFn: () => fetch("/v0/settings").then((r) => r.json()),
+    queryFn: async () => {
+      const response = await fetch("/v0/settings");
+      return (await response.json()) as Settings;
```
laulauland: nit: we could use the `SettingsSchema` here to validate the result as opposed to casting it, at the same time as we're controlling the API it's probably fine

laulauland (approved):
Looks good!

frontend/src/queries/settings.ts
stephlow: Good call! Updated 👍

fpx-lib/src/api/models.rs
```diff
@@ -140,3 +140,32 @@ impl From<SpanAdded> for ServerMessage {
...
+    pub updated_at: time::OffsetDateTime,
+    pub parsed_payload: Span,
+}
+
+impl Into<TypeScriptCompatSpan> for crate::data::models::Span {
+    fn into(self) -> TypeScriptCompatSpan {
+        TypeScriptCompatSpan {
+            span_id: Some(self.span_id.clone()),
+            trace_id: Some(self.trace_id.clone()),
+            created_at: self.start_time.into(),
```
hatchan: Think for both the created_at and updated_at we should just use the same timestamp. Since there really isn't a situation in which we can update a span. I think having both set to the end_time makes the same most sense

hatchan (approved):

fpx-lib/src/api/models.rs
stephlow: Makes sense, will update!

PR #162, opened 2024-08-20
Website workflow update and a button tweak

laulauland:
Just a couple of small tweaks to make this thing work

### DIFF
--- a/.github/workflows/check_website.yaml
+++ b/.github/workflows/check_website.yaml
@@ -9,9 +9,6 @@ on:
 jobs:
   publish:
     runs-on: ubuntu-latest
-    permissions:
-      id-token: write
-      contents: read
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


.github/workflows/check_website.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - "*"
+    paths:
+      - "www/**"
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
```
hatchan: We only need these to assume a AWS role.
```suggestion
```

.github/workflows/publish_website.yaml
```diff
@@ -1,13 +1,18 @@
+name: Build and publish website
```
hatchan, line 1: We could have a workflow with two jobs, 1 to create the site and 1 to deploy that artifact. Saves on a bunch of duplication

hatchan (approved):

.github/workflows/publish_website.yaml
laulauland: yea that's a good idea, bookmarking this for the next release of the website when we update some of the docs content.

PR #166, opened 2024-08-20
Allow compression to be disabled

hatchan:
The wrangler environment already takes care of this, so if we do it in our app then we potentially compress it twice, and thus the client would only see a compressed payload and not the JSON payload


stephlow (approved):
Nice! Thanks for the quick fix

PR #167, opened 2024-08-20
Remove package-lock.json

flenter:
I don't think we need this anymore right?


brettimus:
yes! i think it's easy to accidentally sneak these back in. should we gitignore `package-lock.json`?

brettimus:
i'm going to add to this PR to also gitignore package-lock files

brettimus (approved):

PR #169, opened 2024-08-20
Show service host in history if it is not target service

brettimus:
Also:

- Fix problem with initial load that always set the first url's host to `http://localhost:8787`

- When loading a request from history, only reload query params in the path input if they were defined explicitly in the path. Otherwise, load query params into teh Query Params form in the UI

image


laulauland (approved):
lgtm

PR #173, opened 2024-08-21
If setting is not json, fall back to its og value

brettimus:
things blow up when i try to run the api rn. `getSetting` tries to parse the setting value as json, but sometimes it's just a string

image


laulauland (approved):
💥

PR #174, opened 2024-08-21
Modify OtelSpan schema to handle null attributes

brettimus:
It's possible an attribute on an event could end up being null, and we don't want to break the UI because of it.

This happened with [FP-4019](https://linear.app/fiberplane/issue/FP-4019/otel-middleware-passing-a-function-to-console-serializes-message)

At a certain point we have to be defensive for this case in the UI. So I added `null` as a possible attribute value in the OtelSpanSchema

## Other changes

- Safely parse json when getting all settings from the database
- Always run `db:migrate` when running the ts api in dev

### DIFF
--- a/api/src/lib/settings/index.ts
+++ b/api/src/lib/settings/index.ts
@@ -73,7 +73,9 @@ export async function getAllSettings(
 
   const mappedToSchema = results.reduce<Record<string, string>>(
     (acc, setting) => {
-      acc[setting.key] = setting.value ? JSON.parse(setting.value) : undefined;
+      acc[setting.key] = setting.value
+        ? safeParseJson(setting.value)
+        : undefined;
       return acc;
     },
     {},
--- a/api/package.json
+++ b/api/package.json
@@ -16,7 +16,7 @@
   },
   "types": "./src/lib/types.ts",
   "scripts": {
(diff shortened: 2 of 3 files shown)


api/src/lib/settings/index.ts
```diff
@@ -73,7 +73,9 @@ export async function getAllSettings(
 
   const mappedToSchema = results.reduce<Record<string, string>>(
     (acc, setting) => {
-      acc[setting.key] = setting.value ? JSON.parse(setting.value) : undefined;
+      acc[setting.key] = setting.value
+        ? safeParseJson(setting.value)
+        : undefined;
```
stephlow, line 78: I don't think this is required after running the migration. The migration will clear the settings and the api endpoint for upserting settings validates the data with the `SettingsSchema`

If we have breaking settings changes we should write migrations for those?

stephlow (approved):

api/src/lib/settings/index.ts
brettimus: yeah it's not but also if some non-json sneaks its way into one field in the db, then the entire settings route will crash at this point 😬 

i know defensive coding isn't always the best but! seemed better than the alt

PR #178, opened 2024-08-21
Otel Client Lib: Don't delete those precious headers (Bun/Node)

brettimus:
For some reason the code I put in here fixes and issue that @oscarvz and @Nlea were having with their sample app. 

Headers were getting fully devoured by our instrumentation library... maybe a node/bun specific bug? (Since they're running on bun)


brettimus:
We won't prioritize merging this week, as don't have time to test whether this breaks anything in local workers (wrangler) environments

stephlow (approved):

oscarvz (approved):

PR #179, opened 2024-08-22
Turn on form body types, add Discord link, tweak Fpx Worker settings copy

brettimus:
Title says it all


frontend/src/Layout.tsx
```diff
@@ -47,7 +47,7 @@ export const Layout: React.FC<{ children?: React.ReactNode }> = ({
               <WebhoncBadge />
             </div>
           )}
-          <div className="flex items-center gap-2 border-l">
+          <div className="flex items-center gap-0 border-l px-1">
```
flenter, line 50: Is the gap-0 needed?

flenter (approved):
👍

frontend/src/Layout.tsx
brettimus: probably not BUT IT'S TOO LATE NOW

PR #183, opened 2024-08-23
Node.js support

### LINKED ISSUE #194: Node.js runtime support
## Goal:

support running the client library and Fiberplane Studio with apps running on Node.js runtimes

### Identified issues:
- Client library doesn't pick up on `FPX_ENDPOINT` variable and never activates.
- Requests never reach the application once the client library is activated.

brettimus:
There were several issues that prevented the client library from working with Hono-node apps. 

First, the `FPX_ENDPOINT` wasn't accessed from the environment correctly (for Node).

Second, in attempting to serialize the `env` that's passed to `fetch`, we were serializing an object with a circular reference, which broke all incoming requests. That was an oepsie woepsie.

The solutions thus far:

- Correctly access FPX_ENDPOINT in a Node runtime
- Serialize `process.env` instead of the `env` passed to `app.fetch`, if the latter looks like the wonky env from the hono-node runtime
- Be sure to safely serialize env vars, looking out for any circular references

Finally, there was an issue with how we typed the proxied `fetch` function. We needed to mark `ExecutionContext` as an optional param, since non-Cloudflare envs don't provide it.

## Node.js Test App

Local development testing with the example node app:

```sh
# Shell 1
pnpm i
cd examples/node-api
pnpm dev
```

```sh
# Shell 2
cd api
FPX_WATCH_DIR=../examples/node-api pnpm dev
```

## Documentation

The Get Started docs page has been updated to reflect Node.js support.

This includes the addition of a component for rendering different example code for different platforms.

image

## TODO

- [x] Test changes against Worker apps to look for regressions
- [x] Document Node.js support
- [x] Fix type of proxied `app.fetch`
- [x] Serialize actual env vars
- [x] Investigate why Node.js Hono apps never get to execute their handlers
- [x] Safely serialize the `env`


packages/client-library-otel/src/instrumentation.ts
```diff
@@ -264,3 +266,26 @@ function mergeConfigs(
...
+  const fpxEndpointFromEnv =
+    typeof env === "object" && env !== null
+      ? (env as Record<string, string | null>).FPX_ENDPOINT
+      : null;
+
+  if (fpxEndpointFromEnv) {
+    return fpxEndpointFromEnv;
+  }
+
+  if (typeof process !== "undefined" && typeof process.env !== "undefined") {
```
brettimus, line 286: todo: factor this out into a smarter helper, or do some basic runtime detection like we did in autometrics

brettimus:
We're almost there! Just need some regression testing on Cf apps

brettimus:
Alright, with latest rebasing, this needs some fresh testing on some sample node apps

pietrodev07 (approved):

mies (approved):

PR #185, opened 2024-08-23
Allow spans to be deleted

hatchan:
Allow deleting a specific span or all spans of a trace

NOTE: Support for deleting a batch of spans or having a cronjob that will delete old spans will be added in another PR.

### DIFF
--- a/fpx-lib/src/api/handlers/spans.rs
+++ b/fpx-lib/src/api/handlers/spans.rs
@@ -101,3 +101,40 @@ impl From<DbError> for ApiServerError<SpanListError> {
         ApiServerError::CommonError(CommonError::InternalServerError)
     }
 }
+
+#[tracing::instrument(skip_all)]
+pub async fn span_delete_handler(
+    State(store): State<BoxedStore>,
+    Path((trace_id, span_id)): Path<(String, String)>,
+) -> Result<StatusCode, ApiServerError<SpanDeleteError>> {
+    let tx = store.start_readonly_transaction().await?;
+
+    hex::decode(&trace_id)
+        .map_err(|_| ApiServerError::ServiceError(SpanDeleteError::InvalidTraceId))?;
+    hex::decode(&span_id)
+        .map_err(|_| ApiServerError::ServiceError(SpanDeleteError::InvalidSpanId))?;
+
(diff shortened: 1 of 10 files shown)


fpx-lib/src/api/handlers/spans.rs
```diff
@@ -101,3 +101,40 @@ impl From<DbError> for ApiServerError<SpanListError> {
...
+pub async fn span_delete_handler(
+    State(store): State<BoxedStore>,
+    Path((trace_id, span_id)): Path<(String, String)>,
+) -> Result<StatusCode, ApiServerError<SpanDeleteError>> {
+    let tx = store.start_readonly_transaction().await?;
+
+    hex::decode(&trace_id)
+        .map_err(|_| ApiServerError::ServiceError(SpanDeleteError::InvalidTraceId))?;
+    hex::decode(&span_id)
+        .map_err(|_| ApiServerError::ServiceError(SpanDeleteError::InvalidSpanId))?;
```
mellowagain, line 115: would be cool to write a custom extractor so this wouldnt have to be repeated for every route we use those on

fpx-lib/src/api/handlers/spans.rs
```diff
@@ -101,3 +101,40 @@ impl From<DbError> for ApiServerError<SpanListError> {
...
+
+    let _ = store.span_delete(&tx, &trace_id, &span_id).await?;
```
mellowagain: is this marked `#[must_use]` that this is bound to `_`?

mellowagain (approved):

fpx-lib/src/api/handlers/spans.rs
hatchan: I'll add a type for it, that can do the conversion :+1:

fpx-lib/src/api/handlers/spans.rs
hatchan: No it isn't, will remove the assignment

fpx-lib/src/api/handlers/spans.rs
hatchan: https://linear.app/fiberplane/issue/FP-4039/create-typed-traceidspanid

stephlow (approved):

PR #197, opened 2024-08-26
Add the ability to replay requests from the details view

laulauland:
Addresses FP-3855. This PR adds the ability to replay requests from the request details view.

To achieve this, we're adding 3 hooks:
- `useShouldReplay` which should gate replayability only to json/text requests
- `useReplayRequest` which enables replaying the request
- `useMostRecentRequest` which informs the UI if the currently viewed request is the most recent one

TODO:
- [x] enable a keyboard shortcut cmd-enter to replay the request
- [x] fix up the type errors in `useReplayRequest`
- [x] uncomment the documentation regarding the "Replay" button

Of note: we've taken some shortcuts with regards to showing intermediate states while the request is loading or having some more robust interaction patterns in general. This is because there's explorative work in play that if validated will remove the request detail view in general in which case we shouldn't spend too much time on this.


studio/src/pages/RequestDetailsPage/RequestDetailsPageV2/RequestDetailsPageV2Content.tsx
```diff
@@ -76,48 +107,97 @@ export function RequestDetailsPageContentV2({
...
         </div>
-        {pagination && (
-          <div className="flex gap-2">
+        <div className="flex gap-2 items-center">
+          {!isMostRecentTrace && (
+            <Link
+              className="text-blue-600 pr-4"
+              to={`/requests/otel/${mostRecentTraceId}`}
+            >
+              View most recent response ↗
```
brettimus: this can squish things quite a bit on half a laptop screen (snap ui to half your screen width)

could hide the word `response` except on smaller screens

also could only show the http summary on large screens

image

studio/src/pages/RequestDetailsPage/hooks/useReplayRequest.ts
```diff
@@ -0,0 +1,159 @@
...
+  const filterReplayHeaders = useCallback(
+    (
+      headers: Array<{
+        key: string;
+        value: string;
+        id: string;
+        enabled: boolean;
+      }>,
+    ) => {
+      const excludedHeaders = [
```
brettimus, line 43: nit: could use Set

studio/src/pages/RequestDetailsPage/hooks/useReplayRequest.ts
```diff
@@ -0,0 +1,159 @@
...
+        "x-fpx-trace-id",
+        "transfer-encoding",
+        "content-length",
+        "host",
+        "connection",
+        "accept-encoding",
+        "user-agent",
+        "referer",
+        "origin",
+        "cookie",
```
brettimus, line 53: sure we wanna filter out cookie?

brettimus: if someone attached a cookie for example, this breaks auth.

studio/src/pages/RequestDetailsPage/hooks/useReplayRequest.ts
```diff
@@ -0,0 +1,159 @@
...
+      value,
+      enabled: true,
+    }));
+
+    return filterReplayHeaders(headers);
+  }, [requestHeaders, filterReplayHeaders]);
+
+  const replayBody = useMemo(() => {
+    const body = getRequestBody(span);
+    try {
```
brettimus, line 85: no need to refactor, but for future reference, i made a utility for this kind of logic

utility is called `isJson`, should be importable from the utils module

studio/src/pages/RequestDetailsPage/hooks/useReplayRequest.ts
```diff
@@ -0,0 +1,159 @@
...
+      setActiveResponse,
+    },
+  );
+
+  const replay = useCallback(
+    (e: React.FormEvent<HTMLButtonElement>) => {
+      e.preventDefault();
+      return makeRequest(
+        {
+          addServiceUrlIfBarePath: (replayPath) => replayBaseUrl + replayPath,
```
brettimus, line 122: sorry you had to probably scratch your head to understand then fudge this 😅

brettimus (approved):
some flakiness on getting the "view latest" cta to appear for me, but might be specific to running things locally + a stale websocket

only thing that'd be worth considering imo is how to deal with horizontal space on smaller-ish screens, like half of a laptop screen 

shorter copy, smaller font, something of the like

PR #201, opened 2024-08-28
Starlight workshop

laulauland:
Adds a simple always-draft cheatsheet for using Starlight


brettimus:
image

this looks a little clunky 👀

www/src/main.css
```diff
@@ -124,3 +124,23 @@ h6 {
     text-decoration: underline;
   }
 }
+
+
+/* Styling buttons */
+
+kbd {
```
brettimus: might be good to put these styles on a kbd that has a particular classname? 

then you can revive the `KeyboardButton` component and just make sure it uses `<kbd>` under the hood to remain semantically apropos

brettimus: targeting `kbd` directly has the side effect of styling, for example, the keyboard shortcuts for search (i linked to an image in my comment on this pr)

laulauland: ahhh yea good catch

brettimus (approved):

PR #207, opened 2024-08-28
CLI Fix: Fall back to `cwd` when no project root is found

brettimus:
title says it all

to test:

**this should cause an error**

```sh
cd ~
mkdir meh
cd meh
npx @fiberplane/studio@latest
```

repeat the process with the pkg from this PR, and it should not error 

```sh
pnpx https://pkg.pr.new/fiberplane/fpx/@fiberplane/studio@207
```


brettimus:
tested using the process in the pr overview

stephlow (approved):

oscarvz (approved):

PR #208, opened 2024-08-29
Client library: Use integers in span attributes again

brettimus:
Revert changes from https://github.com/fiberplane/fpx/pull/137/files

Background: The otel rust collector was rejecting certain spans that had integer-values for attributes. We coerced these values to strings to make things work, but it turns out this was a bug in the rust lib.

We need to wait until that bug is resolved and released before updating. Ergo:

**DO NOT MERGE UNTIL OTEL-RS HAS THE FOLLOWING PR MERGED AND RELEASED: https://github.com/open-telemetry/opentelemetry-rust/pull/1906**


brettimus:
it seems like the otel Rust change has been merged upstream! we can merge this alongside changes to our rusty ingestion piece

PR #210, opened 2024-08-29
Make timeline prominent on request builder page

flenter:
image

This PR streamlines the requestor page by adding the timeline to the main response section.

The downside is that some details are now missing on the Request details page (like headers and the query parameter table).  The page still looks ok, but is just missing some data
image

Ideally we'd merge the two pages, so i'm not sure how much of an issue this is and there are plans to remove the requests (details) page in the future so I don't want to spend too much time on this


studio/src/pages/RequestDetailsPage/RequestDetailsPageV2/useOrphanLogs.ts
```diff
@@ -20,7 +20,12 @@ export function useOrphanLogs(traceId: string, spans: Array<OtelSpan>) {
             }
             // TODO - Use a more deterministic ID - preferably string that includes the trace+span+event_index
             const logId = Math.floor(Math.random() * 1000000);
-            const orphanLog = convertEventToOrphanLog(traceId, logId, event);
+            const orphanLog = convertEventToOrphanLog(
+              traceId,
+              logId,
+              event,
+              span.span_id,
```
flenter, line 27: This is a change for the future, where we want to be able to link back to the related spans from log messages.

brettimus:
first note: i would like to be able to see the request headers somehow, even if they're collapsed.

reason is that i used that info to catch a bug in oscar/nele's app last week. a bug i wasn't even looking for per se!

brettimus:
nit: would be nice to get more top padding on this message, but maybe that's a personal thing

image

brettimus:
timeline is pretty squished and unclear out of the box, for a small test request. maybe could re-arrange this to render vertically on smaller screens? ((use case: i have studio on one half of my screen, ide on the other half))

image

brettimus:
another nit: would go for rounded corners on these highlighted areas

image

studio/src/components/Timeline/DetailsList/TimelineDetailsList.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <div
+          key={getId(item)}
+          onMouseEnter={() => setHighlightedSpanId(getId(item))}
+          onMouseLeave={() => setHighlightedSpanId(null)}
+          className={cn(
+            "p-2 max-w-full overflow-hidden",
+            "border-l-2 border-transparent transition-all bg-transparent",
+            "hover:bg-primary/10 hover:border-blue-500",
+            "data-[highlighted=true]:bg-primary/10",
+            "relative after:absolute after:bottom-[-4px] after:bg-muted-foreground/30 after:w-full after:h-px last:after:h-0",
```
brettimus, line 32: nit: not wild about the border around the timeline. might be good to get some ideas from simon on how to tweak this

brettimus:
one last visual thing: decreasing size of log text (it's significantly larger than everything else)

image

mies:
Looks good. One nitpicky thing but would have to see it in action; but I'm usually not a fan of boxes in boxes (the rounded corner 1px rectangle around the timeline)

mies (approved):

brettimus:
gonna merge but we need a quick follow up to address the following:

- hierarchy of logs: do we group under their span? should we have a dedicated log table?
- quickly expanding the response panel to free up space

brettimus:
@mies fwiw i removed the box-in-box

PR #212, opened 2024-08-29
Publish new canary tag of hono otel client lib after merging to main

brettimus:
This PR adds a GitHub action that will build and publish a new `canary` tag of the hono otel client library.

The action should trigger when:

- There's a merge to `main` && 
- The `packages/client-library-otel` code has changed
- The version in `packages/client-library-otel/package.json` has not already been published

## TODO

- [x] Check that we have an NPM token

### DIFF
--- a/.github/workflows/publish_hono_otel_library.yml
+++ b/.github/workflows/publish_hono_otel_library.yml
@@ -1,0 +1,52 @@
+name: Publish Canary Hono Otel Client Lib
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'packages/client-library-otel/**'
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 9
+          run_install: false
+
+      - name: Install dependencies
+        run: pnpm install
+        env:
+          CI: true
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+          cache: "pnpm"
+
+      - name: Check if version is already published
+        id: check_version
+        run: |
+          PACKAGE_VERSION=$(node -p 'require("./packages/client-library-otel/package.json").version')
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_hono_otel_library.yml
```diff
@@ -0,0 +1,52 @@
...
+          PACKAGE_VERSION=$(node -p 'require("./packages/client-library-otel/package.json").version')
+          if npm view @fiberplane/hono-otel@$PACKAGE_VERSION version &> /dev/null; then
+            echo "VERSION_EXISTS=true" >> $GITHUB_OUTPUT
+          else
+            echo "VERSION_EXISTS=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Publish to npm with @canary tag
+        if: steps.check_version.outputs.VERSION_EXISTS == 'false'
+        run: pnpm publish --filter @fiberplane/hono-otel --tag canary --access public
```
hatchan, line 46: Won't it make more sense to have every commit update the canary tag, regardless of the version that is already published? To mean it makes sense to have whatever is in main, be `canary`.

Depending on when the version gets bumped in the package, it could mean that no more `canary` packages will be published if the version is published, but the package is not yet bumped.

brettimus: yeah so i did this a little defensively, since pnpm (locally) doesn't let you publish if you have uncommitted changes. 

i could modify the action to bump the version and push that change, but hesitated bc it'd require wiring up some more stuff presumably 

@keturiosakys thoughts?

hatchan: From the point of view of the CI we should never have any uncommitted changes, right? So it should just always be able to push. 

I would go for less logic and just push it always.

flenter: Arent' there restrictions about publishing a version that's already published?  Ah yes here: https://docs.npmjs.com/policies/unpublish I once ran into issues while i was trying to setup my mono repo with automated publishing :)

hatchan: I guess I didn't realize how this works with NPM. As it seems that a tag requires a version to be published and it kinda acts like a symlink.

hatchan (approved):

PR #214, opened 2024-08-30
Use matching middleware as context in ai req generation

brettimus:
## Overview

This PR adds the ability to inject, as context for AI Request Generation:tm:, the stringified function code of all middleware that matches a given route.

The result is smarter request generation. If certain middleware expects a specific header, or query param, the AI will be able to auto-populate it. For example, the AI will now try to populate the "x-interesting-header" in the following api:

```ts
app.use(async (c, next) => {
  const interestingHeader = c.req.header("x-interesting-header");
  // do something with interestingHeader
  await next();
})

app.get("/", c => c.text("Hello Hono!");
```

Because the frontend only sends along _matching_ middleware as context, the following middleware would not be sent along as context for `GET /`, since the middleware would not fire for the route:

```ts
app.get("/", c => c.text("Hello Hono!");

app.use(async (c, next) => {
  const interestingHeader = c.req.header("x-interesting-header");
  // do something with interestingHeader
  await next();
})
```

### Extra Goodie

When you update your routes with a new route, the order is now reflected in the routes sidebar.

So, if you had:

- `app.get("/", ...)`
- `app.get("/api/stuff", ...)`
- `app.get("/api/things", ...)`

And you updated your routes to

- `app.get("/", ...)`
- `app.get("/api/stuff", ...)`
- `app.post("/api/stuff", ...)`
- `app.get("/api/things", ...)`

Then the routes sidebar will now have the ordering:

- `GET /`
- `GET /api/stuff`
- `POST /api/stuff`
- `GET /api/things`

Instead of

- `GET /`
- `GET /api/stuff`
- `GET /api/things`
- `POST /api/stuff`

## Database Changes

The `app_routes` table needed to be modified as follows:

1. Add the column `registration_order`, which should be the index in the array of routes returned by the app
2. Use an auto-incrementing primary key on the table, in lieu of a computed primary key

### `registration_order`
The `registration_order` column is necessary for understanding which middleware will match for a given path. 
Its default value is `-1`, meaning unregistered.

### Primary Key changes

> **SQLite doesn't allow altering primary keys** so I had to use the trick of creating a shadow copy of the table in order to define a new, identical, table with a different primary key.

We needed a new primary key for the table because the table computed its primary key from the columns:

- method
- path
- handler_type
- route_origin

However, it's possible for multiple middleware to be registered with the same method and path. For example:

```ts
app.use(async (c, next) => {
  console.log("middleware 1");
  await next();
});

app.use(async (c, next) => {
  console.log("middleware 2");
  await next();
});
```

In such a case, both of these middleware will have the column value:

- method: "ALL"
- path: "/*"
- handler_type: "middleware"
- route_origin: "discovered"

So, I opted to ditch the old computed primary key, and handle "collision" logic on the application side (see: API Changes).

This way, we can record all the app's middleware accurately when we receive a route probe.

### `/v0/probed-routes`

Database logic for the `POST /v0/probed-routes` route now lives in `lib/app-routes.ts`.

When we receive a route probe, the logic is now:

1. Mark all current routes in the database as "unregistered"
2. Delete all middleware (we don't want crufty middleware)
3. Re-register all middleware, insert new routes, and upsert any existing (matching) routes

All of this is done in a transaction, but it still suffers from N+1 query problem... which isn't too much of a problem since we're operating locally.

### `inference`

The prompt invocation helper now will format an array of middleware and inject it as context in the prompt.

There is a hack here: We remove `reactRenderer` middleware if we see any.

## Frontend Changes

The `useAi` hook now accepts an array of `matchingMiddleware`.

The matched middleware is computed from the list of routes and middleware. This matching logic is a little involved, but there are comments.

### DIFF
--- a/studio/src/pages/RequestorPage/reducer/reducer.ts
+++ b/studio/src/pages/RequestorPage/reducer/reducer.ts
@@ -3,11 +3,7 @@ import { enforceFormDataTerminalDraftParameter } from "../FormDataForm";
 import type { KeyValueParameter } from "../KeyValueForm";
 import { enforceTerminalDraftParameter } from "../KeyValueForm/hooks";
 import type { ProbedRoute } from "../queries";
-import { findMatchedRoute } from "../routes";
-import {
-  findAllMiddlewareMatches,
-  findSmartRouterMatches,
-} from "../routes/match";
+import { findAllSmartRouterMatches, findMatchedRoute } from "../routes";
 import {
   type RequestMethod,
   type RequestMethodInputValue,
@@ -42,6 +38,7 @@ const _getActiveRoute = (state: RequestorState): ProbedRoute => {
       handler: "",
       handlerType: "route",
       currentlyRegistered: false,
+      registrationOrder: -1,
       routeOrigin: "custom",
       isDraft: true,
     }
@@ -712,10 +709,10 @@ export function useRequestor() {
    * ...
    */
   const getMatchingMiddleware = useCallback(() => {
-    // TODO
     const path = state.path;
     const method = state.method;
     const requestType = state.requestType;
+
     const matchedRoute = findMatchedRoute(
       state.routes,
       removeBaseUrl(state.serviceBaseUrl, path),
@@ -736,7 +733,7 @@ export function useRequestor() {
       (r) => r.handlerType === "middleware",
     );
 
-    const middlewareMatches = findAllMiddlewareMatches(
+    const middlewareMatches = findAllSmartRouterMatches(
       filteredMiddleware,
       removeBaseUrl(state.serviceBaseUrl, path),
       method,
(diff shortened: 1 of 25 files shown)


laulauland (approved):
note: database sometimes fails to update so users might need to blow it away before using this

studio/src/pages/RequestorPage/reducer/reducer.ts
```diff
@@ -683,12 +705,83 @@ export function useRequestor() {
...
+    );
+
+    const middlewareMatches = findAllSmartRouterMatches(
+      filteredMiddleware,
+      removeBaseUrl(state.serviceBaseUrl, path),
+      method,
+      requestType,
+    );
+
+    // console.log("matched route", matchedRoute);
```
flenter: 👋

flenter (approved):

studio/src/pages/RequestorPage/reducer/reducer.ts
brettimus: in the future you can add a code suggestion to remove lines like this! then i can just accept and merge ((i'll try to do the same on pr reviews))

brettimus:
```suggestion
```

PR #216, opened 2024-08-30
TypeScriptCompat* model changes for the Zed extension

laulauland:
A little hackday experiment creating an extension for [Zed](zed.dev) to insert trace details into assistant panel context.

UPDATE: the extension has been prepared and published on the Zed extension registry. It needed some changes to the Rust TypeScriptCompat* models that are detailed in this PR. Note, that the Zed extension refers directly to the `fpx-lib` dep as a git dependency on `zed-extension` branch, so changes there will need to happen once this is merged.

UPDATE 2: This branch willl not be merged. There's work being done to consolidate all TypescriptCompat* structs in a different PR (#180 ). Keeping this branch open only for the initial release of Zed extension that will need to be updated.

~NOTE: it currently makes some slightly terrifying and probably breaking changes to our `fpx-lib` crates so maybe don't merge as is.~

https://github.com/user-attachments/assets/c2f6c79b-4f17-4caa-8709-bd36b1d07bf5


laulauland:
closed in favor of #180

PR #221, opened 2024-09-02
Timeline tweaks

flenter:
Included:
* change layout (use vertical stacking instead of horizontal for request/response panels)
* rename timeline to `logs & events`
* remove duration column in the graph (spans show their duration on hover as well as in the list)
image

### DIFF
--- a/studio/src/pages/RequestorPage/ResponsePanel/ResponsePanel.tsx
+++ b/studio/src/pages/RequestorPage/ResponsePanel/ResponsePanel.tsx
@@ -246,9 +246,9 @@ function ResponseSummary({
   const url = isRequestorActiveResponse(response)
     ? response?.requestUrl
     : parsePathFromRequestUrl(
-      response?.app_requests?.requestUrl ?? "",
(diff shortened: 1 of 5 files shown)


brettimus:
i guess with the added real estate we could put "Response" back in the tabs list? then show the other part of the summary with the 200 next to it

i'll leave it up to you, we can tweak in a follow-on

brettimus (approved):

studio/src/pages/RequestorPage/ResponsePanel/ResponsePanel.tsx
```diff
@@ -182,7 +182,7 @@ export function ResponsePanel({
                         ) : (
                           <CaretRightIcon className="w-4 h-4 cursor-pointer" />
                         )}
-                        Timeline
+                        Logs & events
```
brettimus:
```suggestion
                        Logs & Events
```

capitalizing looks nice to me

stephlow (approved):

oscarvz (approved):

PR #229, opened 2024-09-03
Brett Presents: 2x The Fun! Cf binding errors, and a fix to RegExp router path parameter parsing

brettimus:
Found an unrelated bug while working on rendering Error results from CF bindings and it rolled into this PR bc the fix was kind of involved.

This PR does two very separate things:

1. Will render an `Error` section for CF bindings when the call to a binding resulted in an error 

image

2. (Not screen-shottable) Fixes an issue where the `SmartRouter` returned a match object that mapped path parameters to their _indices in a regexp match object_ instead of their actual values. Turns out we weren't handling that case well because we hadn't actually bumped into it yet.

### DIFF
--- a/studio/src/pages/RequestorPage/routes/match.ts
+++ b/studio/src/pages/RequestorPage/routes/match.ts
@@ -1,3 +1,4 @@
+import type { ParamIndexMap, ParamStash } from "hono/router";
 import { RegExpRouter } from "hono/router/reg-exp-router";
 import { SmartRouter } from "hono/router/smart-router";
 import { TrieRouter } from "hono/router/trie-router";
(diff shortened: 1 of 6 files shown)


studio/src/pages/RequestorPage/routes/match.ts
```diff
@@ -153,14 +156,24 @@ const isMatchResultEmpty = <R extends SmartRouter<T>, T>(
...
+ */
 const unpackMatches = <R extends SmartRouter<T>, T>(
   result: ReturnType<R["match"]>,
 ) => {
   const matches = [];
   if (result.length === 2) {
     for (const m of result[0]) {
       if (m[0] !== undefined) {
-        matches.push(m);
+        const handler = m[0];
```
brettimus, line 169: this is the fix to path parameter parsing!

mies (approved):

PR #230, opened 2024-09-03
Add some Zustand

flenter:
This PR reorganizes things a bit as part of the move to a centralized state management.

I think everything still works, but i'm not an expert on all features/code here so... Please test and see if there's something that's missing (or also good: code that can be removed)


studio/src/pages/RequestorPage/RequestPanel/RequestPanel.tsx
```diff
@@ -22,29 +17,9 @@ import { FileUploadForm } from "./FileUploadForm";
 import { PathParamForm } from "./PathParamForm";
 import "./styles.css";
 import { CodeMirrorJsonEditor } from "@/components/Timeline";
+import { useRequestorStore } from "../store";
 
 type RequestPanelProps = {
-  activeRequestsPanelTab: RequestsPanelTab;
```
brettimus, line 27: noice

brettimus (approved):
approved! didn't notice any regressions locally, and deleted some dead code during review

PR #234, opened 2024-09-04
Improve inputs for headers, query params, to stretch when focused (like HTTPie)

brettimus:
## TODO

- [x] Press `ESC` to blur input
- [x] Investigate why newlines are possible in FireFox
- [x] Test safari.... 🙈 

## Summary

This PR introduces a new input-like component that we can use for adding query/path parameters and headers in their respective forms.

It also factors out the codemirror editor components into their own module in the `components` directory

https://github.com/user-attachments/assets/63fa1608-2458-43bb-8e2d-255efe4a6aa2


brettimus:
image

this is the blocker before merging. something is weird with codemirror and breaking the layout every once in a while

brettimus:
another issue: tab inserts a tab instead of going to next component

brettimus:
Hmmm one annoying thing that came up is that the codemirror inputs swallow our keyboard shortcuts (cmd+g and cmd+b) when they have focus

mies (approved):

laulauland (approved):
Looks good! Ideally we'd "blur" the text input when you hit Escape so that it would unblock using the cmd+b and other keyboard shortcuts.

brettimus:
> Looks good! Ideally we'd "blur" the text input when you hit Escape so that it would unblock using the cmd+b and other keyboard shortcuts.

did it!

studio/src/components/CodeMirrorEditor/CodeMirrorInput.tsx
```diff
@@ -0,0 +1,211 @@
...
+  value?: string;
+  onChange: (value?: string) => void;
+  placeholder?: string;
+  onSubmit?: () => void;
+};
+
+const preventNewlineInFirefox = keymap.of([
+  {
+    key: "Enter",
+    run: (_view) => {
```
flenter: Curious, can this parameter be removed?

studio/src/components/CodeMirrorEditor/CodeMirrorInput.tsx
```diff
@@ -0,0 +1,211 @@
...
+export function CodeMirrorInput(props: CodeMirrorEditorProps) {
+  const { value, onChange, placeholder, width, readOnly, onSubmit } = props;
+
+  const [isFocused, setIsFocused] = useState(false);
+
+  const style = useMemo(
+    () => ({
+      // if specified, the width should be a fixed width, otherwise we take up full width
+      width: width ?? "100%",
+      height: "auto",
```
flenter: Bonus points if the static styling is done via the tailwind classes (which is also done)  and even more points if isFocused can be removed (something like:`focus-within:overlfow-auto`)

flenter (approved):
nice!

studio/src/components/CodeMirrorEditor/CodeMirrorInput.tsx
brettimus: good point — this style object was a holdover from the first attempts at writing a custom component. we can just set the width with a tailwind class that's passed as a prop! refactoring

brettimus: focus might be harder though, since the parent div needs to be aware of focus state, but i will try

brettimus: yeah so focus state needed to remain in order to conditionally load a codemirror extension that helps with truncation. but the styles have been refactored to rely on focus-within!

studio/src/components/CodeMirrorEditor/CodeMirrorInput.tsx
brettimus: indeed it can - was toying with it to see if i had additional info on (e.g.) whether or not "mod" was pressed as well

PR #235, opened 2024-09-04
RequestorPage: on small screen also render bg

flenter:
For the request/response panels

I noticed that on small screens the background was gone and also the selector for content-type was in the wrong place

Before:
after

After: 
after


mies (approved):

oscarvz (approved):

flenter:
This fix is no longer needed (as its already been resolved)

PR #246, opened 2024-09-10
Add key sequence shortcuts

laulauland:
This PR add some navigation flow improvements. It re-adds the custom key sequence hook we used in the initial implementation of keyboard support. 

It also adds the following keyboard shortcuts:
- `G` `L` to toggle logs panel
- `G` `T` to toggle timeline
- `G` `I` to toggle AI test panel
- `G` `R` to toggle focus on routes
- `G` `A` to toggle focus on all requests

TODO:
- [x] add tooltip hovers on each of the elements showing the keyboard shortcut that can be used.


mies (approved):

studio/src/hooks/useKeySequence.ts
```diff
@@ -0,0 +1,89 @@
...
+  useEffect(() => {
+    sequenceRef.current = sequence;
+  }, [sequence]);
+
+  useEffect(() => {
+    callbackRef.current = callback;
+  }, [callback]);
+
+  // create and persist the callback across re-renders
+  const handleKeyPress = useCallback((event: Event) => {
```
flenter: I would prefer the use of `useHandler` instead of `useCallback`. 

By doing so we no longer need to create the `callbackRef` and `sequenceRef` anymore (nor update them).

studio/src/hooks/useKeySequence.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const [ref, setRef] = useState<HTMLElement | null>(null);
+  const [keySequence, setKeySequence] = useState<string[]>([]);
+
+  const timeoutRef = useRef<number | NodeJS.Timeout | undefined>(undefined);
+  const callbackRef = useRef(callback);
+  const sequenceRef = useRef(sequence);
+
+  // ensure that the callback and sequence are updated when they change
+  useEffect(() => {
```
flenter: There's a hook called `useLatest` that we may want to use for this sort of cases. it's very simple, but shows very clearly the intent of the ref:
https://github.com/streamich/react-use/blob/master/src/useLatest.ts

studio/src/hooks/useKeySequence.ts
```diff
@@ -0,0 +1,89 @@
...
+      const updatedSequence = [...prevKeySequence, event.key].slice(
+        -sequenceRef.current.length,
+      );
+
+      return updatedSequence;
+    });
+  }, []);
+
+  useEffect(() => {
+    if (keySequence.join("") === sequenceRef.current.join("")) {
```
flenter: Can you add some comments here. Why are we using a reference for sequence? That should not matter, right? I'm guessing its done because the hook gets called with an array of keys (sequence) and that array may be recreated each render.

flenter (approved):
I added some comments, but nice!

studio/src/hooks/useKeySequence.ts
laulauland: tested it by using the value directly and seems like it works well with it as well.

PR #249, opened 2024-09-11
Toggle and navigate side panel with keyboard

laulauland:
addresses FP-4125

This PR adds the following functionality:
- toggle the sidepanel using mod+b (cmd+b or ctrl+b depending on your OS) at any stage
- Use arrow keys or j/k to navigate up/down the list of routes or requests and hit enter to make a route/request *active*
- Hit / to focus search in the panel
- While searching hit "Enter" to select the first matching item and use arrow keys or j/k to navigate the other matching items
- Hit escape to escape the search input focus, hit escape again to clear the search

There's some duplicated code in how the keyboard logic is handled on Routes and Requests panels but that is due to them being same same but not quite the same implementations (e.g. Route items are `<button>`s and Requests are `<Link>`. An action item for a next refactor here would be to use the same architecture and reuse the logic (ideally: `<Link>` for items and serialize the active route state in the URL in addition to relying on the store)

Also: I renamed some store logic to better reflect keyboard-first needs. "Active" is the most important state (previously: "selected") while "selected" now refers to something a little more shallow (usually selected with a keyboard navigation). This enables us to perform actions on selected but not active routes/requests (like making them active, deleting them etc.)


.github/workflows/build_frontends.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Setup Biome
         uses: biomejs/setup-biome@v2
         with:
-          version: latest
+          version: 1.8.3
```
flenter, line 37: It would be better if we'd use the version that we've specified in the package.json

packages/client-library-otel/package.json
```diff
@@ -47,9 +47,9 @@
     "build:types": "tsc --project tsconfig.json",
     "build:alias": "tsc-alias -p tsconfig.json -f",
     "build:swc": "cd src && swc . -d ../dist --source-maps",
-    "dev": "wrangler dev sample/index.ts",
+    "start-sample": "wrangler dev sample/index.ts",
     "prepublishOnly": "pnpm run build",
     "clean": "rimraf dist",
-    "watch": "nodemon --watch src --ext ts,js,json --exec \"pnpm run build\""
+    "dev": "nodemon --watch src --ext ts,js,json --exec \"pnpm run build\""
```
flenter: How about something like: `build:watch` 

Also: isn't there a development.md file as well?

studio/src/pages/RequestorPage/NavigationPanel/NavigationPanelContent.tsx
flenter, line 1: 👋

studio/src/pages/RequestorPage/NavigationPanel/RequestsPanel/RequestsPanel.tsx
```diff
@@ -22,21 +29,111 @@ export function RequestsPanel() {
...
   });
 
+  useHotkeys(
+    ["Escape", "Enter"],
+    (event) => {
+      switch (event.key) {
+        case "Enter": {
+          if (isInputFocused && filteredItems.length > 0) {
+            setSelectedIndex(0);
+            const firstItemElement = document.getElementById("#item-0");
```
flenter: You probably mean "item-0" (this also applies to "#route-0" further down

studio/src/pages/RequestorPage/Tabs.tsx
```diff
@@ -54,6 +54,7 @@ export const CustomTabsContent = React.forwardRef<
   <TabsContent
     ref={ref}
     {...props}
+    tabIndex={-1}
```
flenter, line 59: Can you add a comment why this is needed?

flenter (approved):
Nice, left a few comments but looks good already

packages/client-library-otel/package.json
laulauland: oh this snuck in from an unrelated thing I was trying. Gonna just revert it

mies (approved):

PR #251, opened 2024-09-13
Update landing page

oscarvz:
Introduces `Hero.astro` with updated styles specifically for the landing page.

We might need to update typography altogether, but it's looking pretty sweet already ✨ 

---
https://github.com/user-attachments/assets/9f28c898-aac2-4b10-876b-524d1e254f8a


laulauland (approved):
🌯🌯🌯

PR #253, opened 2024-09-16
Add support for iterator 

flenter:
Previoulsy iterators would only show up as a very fast function call., but now the generator function's duration will be as long as the iterator takes to be consumed/error out

Also: 
* contains a fix for streaming responses: Streaming responses wouldn't return a result to the client
* There's a new event that signifies when the response is returned:
``` ts
                span.addEvent("first-response");
````
We don't do anything with this in our UI but probably want to (in the future)
* `measure` now has support for  `endSpanManually` (`boolean`). Which allows onSuccess functions to manually call `span.end()` whenever it is deemed appropriate. This is useful for streaming responses. Where the `end` of the span is when the last piece of data has been send.

Finally: the resulting timeline looks like this:
image

### DIFF
--- a/packages/client-library-otel/src/utils/request.ts
+++ b/packages/client-library-otel/src/utils/request.ts
@@ -240,10 +240,33 @@ async function tryGetResponseBodyAsText(
   }
 
   try {
-    return await response.text();
+    if (response.body) {
+      return await streamToString(response.body as ReadableStream);
+    }
   } catch {
-    return null;
+    // swallow error
+    // } catch (error) {
+    // console.error("Failed to read response.body", error);
   }
+
+  return null;
+}
+
+// Helper function to convert a ReadableStream to a string
+async function streamToString(stream: ReadableStream) {
+  const reader = stream.getReader();
+  const decoder = new TextDecoder();
+  let result = "";
+
+  while (true) {
+    const { done, value } = await reader.read();
+    result += decoder.decode(value, { stream: true });
+    if (done) {
+      break;
+    }
+  }
+
+  return result;
 }
 
 export async function getResponseAttributes(
@@ -255,6 +278,7 @@ export async function getResponseAttributes(
   };
 
   const responseText = await tryGetResponseBodyAsText(response);
+
   if (responseText) {
     attributes[FPX_RESPONSE_BODY] = responseText;
   }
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Review by laulauland — APPROVED** [review]

**Comment by brettimus** [comment]
added linear ticket for `"first-response"` UI: https://linear.app/fiberplane/issue/FP-4161/show-first-response-event-for-streaming-responses

File: packages/client-library-otel/src/utils/request.ts [hunk 1]
```diff
@@ -240,10 +240,33 @@ async function tryGetResponseBodyAsText(
...
 
   try {
-    return await response.text();
+    if (response.body) {
+      return await streamToString(response.body as ReadableStream);
+    }
   } catch {
-    return null;
+    // swallow error
+    // } catch (error) {
```
brettimus, line 248: rogue commented-out code

studio/src/components/Timeline/hooks/useAsWaterfall.ts
```diff
@@ -39,6 +39,17 @@ export function useAsWaterfall(
         if ("span" in b && b?.span?.name === "request") {
           return 1;
         }
+
+        // If the time stamp is the same, sort on span_id/parent_span_id
+        // TODO: improve further sorting.
+        if ("span" in a && "span" in b) {
```
brettimus, line 45: ah nice — i saw on monday that with goose quotes, sometimes the first child span rendered above the root request span, forgot to file a ticket. guessing this fixes that!

brettimus:
you could bump the version in the `package.json` for the client library, then we'll automatically create a new canary from this version!

brettimus:
> you could bump the version in the `package.json` for the client library, then we'll automatically create a new canary from this version!

i'm going to do this real quick!

packages/client-library-otel/src/utils/request.ts
```diff
@@ -240,10 +240,33 @@ async function tryGetResponseBodyAsText(
...
+    // console.error("Failed to read response.body", error);
```
brettimus: rogue commented out code 👀 

```suggestion
```

brettimus (approved):

packages/client-library-otel/src/utils/request.ts
flenter: I wasn't ready yet to remove that line... I think i'm ready now... :)

PR #263, opened 2024-09-18
Implement API manager

hatchan:
This allows fpx-app to startup a API whenever a workspace is opened, and stop the API when the workspace is closed.

There are currently two implementations: One that uses the current TS API and starts it by starting processes and managing these. There is also a implementation that uses the Rust API which get started as part of the fpx-app process itself.

Currently the TS implementation is used by default. The Rust API can be used by building the fpx-app with the `fpx-api` feature flag. Example: `cargo tauri dev -f fpx-api`.

Other changes: 
- Move ServerEvents and LibSQLStore from `fpx-cli` to `fpx`
- Fix issue with duplicate recent workspaces 
- Small renames

Resolves: FP-4177, FP-4184


mellowagain (approved):
still got todos in there

hatchan:
> still got todos in there

I've removed the TODO's, but they are still applicable. They are part of another issue. Thanks for the headsup

PR #265, opened 2024-09-18
Taurification

hatchan:
This should contain all the work related to creating a Tauri app.

**Important: This should not be merged yet.**


api/src/app.ts
```diff
@@ -21,6 +22,13 @@ export function createApp(
 ) {
   const app = new Hono<{ Bindings: Bindings; Variables: Variables }>();
 
+  app.use(
```
flenter, line 25: I think this middleware is included in several other places in our routes, so i guess it can be removed there.

fpx/src/api/errors.rs
```diff
@@ -84,10 +84,9 @@ pub enum ApiClientError<E> {
     #[error("An invalid URL was provided: {0}")]
     ParseError(#[from] url::ParseError),
 
-    /// An error occurred in reqwest.
-    #[error("An error occurred while making the request: {0}")]
-    ClientError(#[from] reqwest::Error),
-
+    // /// An error occurred in reqwest.
+    // #[error("An error occurred while making the request: {0}")]
```
flenter, line 88: Can go?

.github/workflows/build_app.yml
```diff
@@ -0,0 +1,98 @@
...
+  # Run on every push to main
+  push:
+    branches: ["main", "tauri-main"]
+
+jobs:
+  build-app:
+    name: Create binary for ${{ matrix.target }}
+    runs-on: ${{ matrix.os }}
+
+    # permissions:
```
flenter, line 21: Can go?

fpx-cli/Cargo.toml
```diff
@@ -0,0 +1,76 @@
+[package]
+name = "fpx-cli"
+version = { workspace = true }
+edition = "2021"
+authors = { workspace = true }
+license = { workspace = true }
+repository = { workspace = true }
+
+[features]
+embed-studio = [] # When enabled it will embed Studio from frontend/dist
```
flenter, line 10: Silly question maybe, but how do you enable it?

studio/src/components/RuntimeProvider/TauriRuntime/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  return await openWorkspace(selected);
+}
+
+async function handleDirectorySelection() {
+  const selected = await open({
+    directory: true,
+    multiple: false,
+    defaultPath: await appDataDir(),
```
flenter, line 42: nitpick, but i would move this await to above the `open` call.

studio/src/pages/RequestorPage/ai/summarize-traces.ts
```diff
@@ -150,9 +155,11 @@ export function serializeTraceForLLM(trace: Array<OtelSpan>) {
 }
 
 export function useSummarizeError(trace?: Array<OtelSpan>) {
+  const apiBaseUrl = useApiBaseUrl();
+
   return useQuery({
     queryKey: ["summarizeError"],
-    queryFn: () => summarizeError(trace),
+    queryFn: () => summarizeError(apiBaseUrl, trace),
```
flenter, line 162: Shouldn't the `queryKey` actually contain something like the traceId to avoid it being cached incorrectly?

flenter (approved):

fpx-cli/Cargo.toml
hatchan: Currently this is not working anymore, so we might just get rid of it. Because this features requires that studio would have been build (ie. the `dist` folder needed to exists) I wanted to make it a optin feature (so that everyone can easily use the fpx-cli). 

There are two situations where you'd use a Rust feature: either in your `Cargo.toml`, where along a dependency you'd ask for a feature. But in this case you'd use something like: `cargo build -F embed-studio` to build the cli with the studio embedded in there. 

For another example, there is the fpx-app, which allows for the TS or Rust API to be enable using a flag.

PR #273, opened 2024-09-23
Panel tweaks

flenter:
This PR merges several panels into a single bottom panel (with tabs). It also:
* maintains the current trace in the URL, so after making a request, refreshing the page no longer clears all the data in the form
* fixes the notification indicator for logs
* fixes some edge cases where traceid wasn't always picked up by all components in the UI

Demo:

https://github.com/user-attachments/assets/afaeeeb7-6a9c-449e-b304-549d1293114c


laulauland (approved):
looks good! Could we add the changelog note plss

stephlow (approved):

PR #283, opened 2024-09-25
Add Neon DB Query event to log table

brettimus:
This PR adds an entry to the log table for Neon DB calls (success/failure).

Upon expansion of the row, you can see quick stats about the call. 

## Screenshots

**Successful query**

image

**Successful query -- expanded**

image

**Erroneous query**

image

**Erroneous query -- expanded**

image

## Improvements (for the future)

- The all red background is a little too intense for me imo
- Newline heavy sql query takes up too much vertical space
- Show results in a table when possible
- Show size of results
- Show connection string (redacted)
- Link to the full span somehow? in the timeline?

### DIFF
--- a/studio/src/pages/RequestorPage/LogsTable/LogsTableRow/NeonEventRow.tsx
+++ b/studio/src/pages/RequestorPage/LogsTable/LogsTableRow/NeonEventRow.tsx
@@ -9,7 +9,7 @@ import {
   TooltipTrigger,
 } from "@/components/ui/tooltip";
 import { useCopyToClipboard } from "@/hooks";
-import { cn, noop } from "@/utils";
+import { cn, getString, noop } from "@/utils";
 import { CopyIcon } from "@radix-ui/react-icons";
 import { useMemo, useState } from "react";
 import type { NeonEvent } from "../types";
@@ -24,7 +24,11 @@ export function NeonEventRow({ log }: { log: NeonEvent }) {
   const { isCopied: isMessageCopied, copyToClipboard: copyMessageToClipboard } =
     useCopyToClipboard();
 
-  const message = `Neon DB Call took ${log.duration}ms`;
+  const isError = !!log.errors?.length;
+
+  const message = isError
+    ? "Neon DB Call failed"
+    : `Neon DB Call took ${log.duration}ms`;
 
   const queryValue = useFormattedNeonQuery(log.sql);
   // This will show `SELECT`, `INSERT`, etc. instead of the full query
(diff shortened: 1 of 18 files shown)


studio/src/pages/RequestorPage/LogsTable/LogsTableRow/NeonEventRow.tsx
```diff
@@ -0,0 +1,134 @@
...
+} from "@/components/ui/tooltip";
+import { useCopyToClipboard } from "@/hooks";
+import { cn, getString, noop } from "@/utils";
+import { CopyIcon } from "@radix-ui/react-icons";
+import { useMemo, useState } from "react";
+import type { NeonEvent } from "../types";
+import { formatTimestamp } from "./shared";
+
+export function NeonEventRow({ log }: { log: NeonEvent }) {
+  // const textColor = "text-gray-500";
```
flenter: 👋

studio/src/pages/RequestorPage/LogsTable/LogsTableRow/NeonEventRow.tsx
```diff
@@ -0,0 +1,134 @@
...
+    : `Neon DB Call took ${log.duration}ms`;
+
+  const bgColor = isError ? "bg-red-500/10" : "bg-green-500/10";
+
+  const queryValue = useFormattedNeonQuery(log.sql);
+  // This will show `SELECT`, `INSERT`, etc. instead of the full query
+  const queryTypePreview = useMemo(() => {
+    const queryType = (queryValue.split("\n")?.[0] ?? "").toUpperCase();
+    return queryType.length < queryValue.length ? `${queryType}...` : queryType;
+  }, [queryValue]);
```
flenter: I like the simple approach, but does this always work? When thinking about showing the type of the query i was thinking about using the response from neon (which contains the type of the SQL statement), but I guess this works assuming that drizzle always formats the query the same way (or perhaps it's not that bad if more than just `SELECT` is being returned here.

flenter (approved):

studio/src/pages/RequestorPage/LogsTable/LogsTableRow/NeonEventRow.tsx
brettimus: 👀

studio/src/pages/RequestorPage/LogsTable/LogsTableRow/NeonEventRow.tsx
brettimus: so it works because of the sql formatter! but i like your solution better, didn't realize it was being returned by the neon api. fixed in latest commit!

PR #286, opened 2024-09-30
Allow form submissions from Headers inputs, Params inputs, JSON Body Input, and Text Body Input

brettimus:
## Overview

Currently, in the UI, when you have focus inside the following inputs, you cannot submit the request with CMD+ENTER:

- Query Params
- Path Params
- Headers
- Body

Now, CMD+Enter should work for all of those.

https://github.com/user-attachments/assets/3c951b88-04cd-46ff-9833-c65be2332f98

***

## Other Improvements

- Fixed a bug where changing the Method input would not update the tabs in the request panel
- Added placeholder to the Query/Header inputs that indicate whether they are for headers or query params, etc
- Implemented the new fancy inputs for the FormData forms

### DIFF
--- a/studio/src/pages/RequestorPage/store/slices/requestResponseSlice.ts
+++ b/studio/src/pages/RequestorPage/store/slices/requestResponseSlice.ts
@@ -13,6 +13,7 @@ import {
   removeBaseUrl,
 } from "../utils";
 import type { RequestResponseSlice, Store } from "./types";
+import { getVisibleRequestPanelTabs } from "../tabs";
 
 export const requestResponseSlice: StateCreator<
   Store,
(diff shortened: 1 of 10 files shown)


studio/src/pages/RequestorPage/store/slices/requestResponseSlice.ts
```diff
@@ -73,6 +74,19 @@ export const requestResponseSlice: StateCreator<
       // Update other state properties based on the new method and request type
       // (e.g., activeRoute, visibleRequestsPanelTabs, activeRequestsPanelTab, etc.)
       // You might want to move some of this logic to separate functions or slices
+
+      // Update visibleRequestsPanelTabs based on the new method and request type
+      state.visibleRequestsPanelTabs = getVisibleRequestPanelTabs({
```
brettimus, line 79: this was a bugfix separate from the main PR

flenter (approved):

PR #290, opened 2024-10-01
Rename Requestornator to ProxiedRequestResponse

brettimus:
We had a nonsense type called `Requestornator`, which I named very fast a few months ago. I've updated the type name to something a bit more descriptive

This type represents an api request (and its response data) that was proxied through our api

The new name is `ProxiedRequestResponse`


flenter (approved):

flenter:
To be honest, i would have preferred `ProxiedRequest` as it is a bit shorter

stephlow (approved):

oscarvz (approved):

PR #291, opened 2024-10-01
Static analysis utilties to improve ai request generation

brettimus:
## Overview

This is a big one. Too big, one might say. But the end result is farrrr more robust ai request generation, which is nice.

I have a README in the new folder in `api/src/lib/expand-function` that can give an idea of how to start testing this locally.

- Pros: Works on test projects, as well as Oscar and Nele's hono-github-tracker (if you modify it to run with wrangler)
- Cons: LOTS of new dependencies, needs to spawn a typescript language server via `npx`, can be slow, and only works for Wrangler projects 😢  for now

### The Problem

Right now, when we need to generate a request with AI, all we have at our disposal are the stringified versions of compiled handler and middleware functions.

If those functions rely on helper utilities to, say, validate certain headers or parameters, we end up in a pickle. The LLM doesn't have enough context to give a valid request back to us. It's effectively blind to certain critical logic.

### The Solution

So, our job is to locate the handler and middleware definitions in the code, and then statically analyze them, to find the definitions of any constants or utilities that they use to execute an incoming request.

To achieve this, we have to do some terrible things.

1. Map the compiled functions back to their locations in the source
2. Parse and analyze the abstract syntax tree of the source function
3. Apply heuristics to identify out-of-scope identifiers in the source function
4. Traverse the codebase to find the definitions of said out-of-scope identifiers
5. Repeat steps (3) and (4) recursively on any utility functions that are used in the functions we analyze

The rest of the pr description will go through the decisions made to implement each of these steps.

The 3rd party libraries/tools at our disposal are:

- `acorn` and `source-map` - libraries for parsing source maps to go from compiled code back to the source
- `typescript` - the typescript compiler api, for parsing and traversing ASTs
- `typescript-language-server`  - wrapper around `tsserver` (with which we communicate over stdio)
- `vscode-jsonrpc` - helpers for interfacing with the ts language server

## Mapping compiled code back to source code

This happens outside the newly introduced `lib/expand-function`. You can find the code (and tests) in `lib/find-source-function`.

There is some ugliness here because at the last moment, I needed to do some optimizations and make this library capable of batch lookups. Definitely room for improvement there.

## Determining out-of-scope identifiers in the function body

`lib/expand-function/identifier-analyzer.ts`

This has unit tests, which in a certain way document the rough edges with this logic.

Check the docstring for more info on the `analyzeOutOfScopeIdentifiers` function. It is somewhat thorough.

## Finding definitions of said out-of-scope identifiers

Alright. This is the most shameful part.

We interface with a locally spawned typescript language server in order to look up definitions of identifiers from the previous step. This requires an interplay between LSP commands and the Typescript Compiler API library (`ts`) in order to look up definitions and stringify them. 

@actualwitch has suggested that we could get rid of this language server dependency! But I got too busy figuring out the source map stuff this week to integrate it in this PR. See: https://github.com/fiberplane/fpx/commit/c477a0986660497201c359afc7540970b6adada3

## How did you test this?

Aside from some light unit and integration tests, I tested on three other projects:

1. `examples/test-static-analysis`
7. `examples/goose-quotes`
8. A forked version of `hono-github-tracker` that uses wrangler instead of vite (https://github.com/brettimus/hono-github-tracker/tree/use-wrangler)

This allowed for testing:

- handlers with lots of middleware, both of which had many helper functions
- handlers that rely on 3rd party code

## A few extra notes

- First request against a route will always be the slowest for static analysis. But once we've done analysis on a route, things will speed up. (3s => 400ms in one test for me)

## TODO

- [x] Expand context of middleware
- [x] Test on one of our bigger sample apps!


api/src/lib/expand-function/README.md
```diff
@@ -0,0 +1,79 @@
+# Static Analysis to Expand Functions
```
brettimus, line 1: TODO - Update this readme

api/package.json
```diff
@@ -56,6 +58,10 @@
     "minimatch": "^10.0.1",
     "openai": "^4.47.1",
     "source-map": "^0.7.4",
+    "typescript": "^5.5.4",
```
flenter, line 61: 👍 this should have been a dependency already

api/src/lib/ai/openai.ts
```diff
@@ -46,19 +50,23 @@ export async function generateRequestWithOpenAI({
     `persona: ${persona}`,
     `method: ${method}`,
     `path: ${path}`,
-    `handler: ${handler}`,
-    `openApiSpec: ${openApiSpec}`,
-    `middleware: ${middleware}`,
+    // `handler: ${handler}`,
```
flenter, line 53: Can this go away? (same applies for the anthropic code)

api/src/lib/expand-function/ast-helpers/ast-helpers.ts
```diff
@@ -0,0 +1,152 @@
...
+export function getDefinitionText(
+  node: ts.Node,
+  sourceFile: ts.SourceFile,
+): DefinitionText | undefined {
+  // Check: Variable declaration with initializer
+  // `const y = ...`
+  // `let x = ...`
+  if (ts.isVariableDeclaration(node) && node.initializer) {
+    return {
+      type: "unknown", // TODO
```
flenter: Can you add a bit more information here?

api/src/lib/expand-function/ast-helpers/ast-helpers.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  // Check if the node is an identifier and the parent node is a variable declaration with initializer
+  // `const myVariable = someValue;`
+  if (
+    ts.isIdentifier(node) &&
+    ts.isVariableDeclaration(node.parent) &&
+    node.parent.initializer
+  ) {
+    return {
+      type: "unknown", // TODO
```
flenter: Same here, what's the thing that should be done in the future?

api/src/lib/expand-function/ast-helpers/ast-helpers.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  return undefined;
+}
+
+export function findNodeAtPosition(
+  sourceFile: ts.SourceFile,
+  position: { line: number; character: number },
+): ts.Node | undefined {
+  function find(node: ts.Node): ts.Node | undefined {
```
flenter, line 104: Couldn't `findNodeAtPosition` simply call itself again? That to me seems like a simpler solution compared to the current nested `find` function.

api/src/lib/expand-function/expand-function.ts
```diff
@@ -0,0 +1,421 @@
...
+  identifiers: OutOfScopeIdentifier[],
+  options: { debug?: boolean } = { debug: false },
+): Promise<ExpandedFunctionContext> {
+  const { debug } = options;
+  const context: ExpandedFunctionContext = [];
+
+  if (!identifiers?.length) {
+    if (debug) {
+      logger.debug(
+        "[debug] No out of scope identifiers found in function, skipping context extraction",
```
flenter, line 176: In the other log call in this file you have the function name in the brackets, would it not be better to do this here as well? Or perhaps it should be in addition to `[debug]`

api/src/lib/expand-function/expand-function.ts
```diff
@@ -0,0 +1,421 @@
...
+            if (!functionNode) {
+              logger.warn(
+                `[extractContext] No function body found for ${identifier.name}`,
+              );
+              continue;
+            }
+
+            // Do type narrowing on the result to appease the call to analyzeOutOfScopeIdentifiers
+            if (
+              !ts.isFunctionDeclaration(functionNode) &&
```
flenter, line 373: I've seen this code elsewhere as well. Would it be handy to have a util function that says: `isFunctionLike`, which checks for all 3 things?

api/src/lib/expand-function/imports/context-for-import.ts
```diff
@@ -0,0 +1,193 @@
...
+      );
+      if (namedImport) {
+        identifierToFind =
+          namedImport.propertyName?.text || namedImport.name.text;
+      }
+    }
+
+    if (identifierToFind) {
+      logger.debug(
+        `[debug][followImport] Identifier to find in file we're importing from: ${identifierToFind}`,
```
flenter, line 133: Nitpick: maybe add a space between `[debug]` and `[followImport]` or perhaps other places should be updated and not have the space? I'm seeing both styles in this pr.

api/src/lib/expand-function/imports/extract-package-name.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function extractPackageName(uri: string): string | null {
+  try {
+    const url = new URL(uri);
+    const path = decodeURIComponent(url.pathname); // Decode URI components
+
+    const nodeModulesIndex = path.indexOf("node_modules");
+
+    if (nodeModulesIndex === -1) {
+      return null;
```
flenter, line 11: If people are using yarn , then there's not always a `node_modules` folder. Won't that be an issue?

api/src/lib/expand-function/search-function/search-compiled-function.ts
```diff
@@ -0,0 +1,122 @@
...
+  const jsFilePath = path.join(compiledJavascriptPath, "index.js");
+  const mapFile = `${jsFilePath}.map`;
+  const sourceMapContent = JSON.parse(
+    await fs.promises.readFile(mapFile, { encoding: "utf8" }),
+  );
+  const jsFileContents = await fs.promises.readFile(jsFilePath, {
+    encoding: "utf8",
+  });
+
+  // const truncatedFunctionString = functionString.slice(0, 100);
```
flenter, line 36: 👋

api/src/lib/expand-function/search-function/search-compiled-function.ts
```diff
@@ -0,0 +1,122 @@
...
+import { findSourceFunctions } from "../../find-source-function/index.js";
+
+/**
+ * Retrieves the source function text from the compiled JavaScript directory.
+ *
+ * On even a medium sized codebase, these lookups can take 600ms
+ *
+ * @param {string} projectPath - The path to the project directory.
+ * @param {string} functionString - The text of the function to retrieve.
+ * @returns {Promise<string | null>} The source function text if found, otherwise null.
```
flenter, line 13: Oops, this seems to be out of date

api/src/lib/expand-function/search-function/search-compiled-function.ts
```diff
@@ -0,0 +1,122 @@
...
+ */
+export async function getSourceFunctionText(
+  projectPath: string,
+  functionString: string,
+): Promise<{ text: string | null; sourceFile: string | null } | null> {
```
flenter, line 18: Why allow for both a `null` and also possibly a `{ text: null, sourceFile: null}` value?

api/src/lib/expand-function/tests/expand-function.test.ts
```diff
@@ -0,0 +1,367 @@
...
+          expect.objectContaining({
+            name: "log",
+          }),
+        ]),
+      );
+
+      // HACK - `log` maps back to workers-types (cloudflare pacakge), but not a `.d.ts` file...
+      //         so it's really difficult to detect that it's actually part of the runtime
+      //         so we just check that we correctly identify the package name and move on for now
+      // expect(result?.context).toEqual(
```
flenter, line 224: Can go?

api/src/lib/expand-function/tests/expand-function.test.ts
```diff
@@ -0,0 +1,367 @@
...
+            definition: expect.objectContaining({
+              text: `sqliteTable("stuff", {
+  id: integer("id", { mode: "number" }).primaryKey(),
+  foo: text(\"foo\").notNull(),
+  createdAt: text("created_at").notNull().default(sql\`(CURRENT_TIMESTAMP)\`),
+  updatedAt: text("updated_at").notNull().default(sql\`(CURRENT_TIMESTAMP)\`),
+})`.trim(),
+            }),
+            // TODO - Follow definition of sqliteTable and sql
+            // context: expect.arrayContaining([
```
flenter, line 284: Can go? (same applies elsewhere)

api/src/lib/expand-function/tsserver/commands.ts
```diff
@@ -0,0 +1,176 @@
...
+export async function openFile(
+  connection: MessageConnection,
+  filePath: string,
+) {
+  const fileUri = getFileUri(filePath);
+  const fileContent = fs.readFileSync(filePath, "utf-8");
+
+  const existingFile = openedFiles.get(filePath);
+  if (existingFile && existingFile.content === fileContent) {
+    // logger.debug("[debug] [openFile] File already opened:", filePath);
```
flenter, line 20: 👋

api/src/lib/expand-function/tsserver/commands.ts
```diff
@@ -0,0 +1,176 @@
...
+  if (!existingFile) {
+    logger.warn(
+      "[warn] Attempting to update a file that wasn't opened:",
+      filePath,
+    );
+    return await openFile(connection, filePath);
+  }
+
+  if (existingFile.content === newContent) {
+    // logger.debug("[debug] [updateFile] File content unchanged:", filePath);
```
flenter, line 54: 👋

api/src/lib/expand-function/tsserver/server.ts
```diff
@@ -0,0 +1,175 @@
...
+ * If an instance already exists, it will be reused to avoid unnecessary initialization.
+ *
+ * @param {string} pathToProject - The path to the project directory.
+ * @returns {Promise<{ connection: MessageConnection; tsServer: ChildProcessWithoutNullStreams }>} The TypeScript language server instance.
+ */
+export async function getTSServer(pathToProject: string) {
+  const resolvedPath = resolve(pathToProject);
+
+  if (tsServerInstance) {
+    // logger.debug(
```
flenter, line 40: 👋

api/src/lib/expand-function/tsserver/server.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  try {
+    const rootUri = getFileUri(pathToProject);
+    logger.debug(
+      chalk.dim(
+        `[debug] Initializing typescript language server with rootUri: ${rootUri}`,
+      ),
+    );
+
+    const _response = await connection.sendRequest("initialize", {
```
flenter, line 107: This seems to be an unused variable. Is this code being linted?

api/src/lib/expand-function/imports/context-for-import.ts
```diff
@@ -0,0 +1,193 @@
...
+      return (
+        modifiers?.some(
+          (modifier) => modifier.kind === ts.SyntaxKind.ExportKeyword,
+        ) &&
+        (ts.isFunctionDeclaration(statement)
+          ? statement.name?.text === identifierName
+          : statement.declarationList.declarations[0].name.getText() ===
+            identifierName)
+      );
+    }
```
flenter, line 190: This is quite dense code to read, how about splitting it up a bit? Something like:

```suggestion
      const hasExportModifier = modifiers?.some(
          (modifier) => modifier.kind === ts.SyntaxKind.ExportKeyword,
        );

      const matchesIdentifier =  (ts.isFunctionDeclaration(statement)
          ? statement.name?.text === identifierName
          : statement.declarationList.declarations[0].name.getText() ===
            identifierName);
            
      return hasExportModifier && matchesIdentifier;
    }
```

api/src/lib/find-source-function/find-source-function.ts
```diff
@@ -0,0 +1,295 @@
...
+      locations: true,
+      sourceType: "module",
+    });
+
+    walkSimple(ast, {
+      FunctionDeclaration(node) {
+        for (const key in functionMap) {
+          if (functionMap[key].foundLocation) {
+            continue;
+          }
```
flenter, line 62: coding style nitpick but i would like it if there's a newline after this statement (also applies elsewhere).

api/src/lib/find-source-function/find-source-function.ts
```diff
@@ -0,0 +1,295 @@
...
+    const ast = parse(jsFileContents, {
+      ecmaVersion: "latest",
+      locations: true,
+      sourceType: "module",
+    });
+
+    walkSimple(ast, {
+      FunctionDeclaration(node) {
+        for (const key in functionMap) {
+          if (functionMap[key].foundLocation) {
```
flenter, line 60: I would create a `value` variable to store `functionMap[key]`, that it's obvious that we're accessing (and even updating it). 

note: whatever we do here, this also applies a couple times further down this file.

api/src/lib/find-source-function/find-source-function.ts
```diff
@@ -0,0 +1,295 @@
...
+  column: number,
+) {
+  return await SourceMapConsumer.with(sourceMapContent, null, (consumer) => {
+    const pos = consumer.originalPositionFor({
+      line, // Line number from JS file
+      column, // Column number from JS file
+    });
+
+    consumer.destroy();
+    // console.log('Original Source:', pos);
```
flenter: 👋

api/src/lib/find-source-function/find-source-function.ts
```diff
@@ -0,0 +1,295 @@
...
+    // Optional: Display the source code snippet if needed
+
+    const returnNullOnMissing = true;
+    const sourceContent = consumer.sourceContentFor(
+      pos.source ?? "",
+      returnNullOnMissing,
+    );
+
+    // console.log('Source Content:\n', sourceContent);
```
flenter: 👋

api/src/lib/find-source-function/find-source-function.ts
```diff
@@ -0,0 +1,295 @@
...
+/**
+ * Finds the original source functions corresponding to compiled function texts using source maps.
+ *
+ * @param {string} jsFilePath - The path to the compiled JavaScript file.
+ * @param {string | Array<string>} compiledFunctionText - The compiled function text(s) to locate in the source.
+ * @param {boolean} [returnNullOnMissing=false] - Whether to return null when a source function is missing.
+ * @param {Object} [hints={}] - Optional hints to provide preloaded source map content or JavaScript file contents.
+ * @param {RawSourceMap | RawIndexMap} [hints.sourceMapContent] - Preloaded source map content.
+ * @param {string} [hints.jsFileContents] - Preloaded JavaScript file contents.
+ * @returns {Promise<FindSourceFunctionsResult>} A promise that resolves to an array of results mapping compiled functions to their sources.
```
flenter: oops, this seems to be no longer current

api/src/lib/find-source-function/find-source-function.ts
```diff
@@ -0,0 +1,295 @@
...
+   * @returns {Promise<{ sourceFunction: string | null; source: string | null } | null>}
+   *          A promise that resolves to the original source function and its source file, or null if not found.
+   */
+  async function lookUpLocation(loc: FunctionLocation) {
+    const functionStartLine = loc?.startLine ?? 0;
+    const functionStartColumn = loc?.startColumn ?? 0;
+    const functionEndLine = loc?.endLine ?? 0;
+    const functionEndColumn = loc?.endColumn ?? 0;
+
+    const [sourceFunctionStart, sourceFunctionEnd] = await Promise.all([
```
flenter, line 245: Can you add a comment that this is done like this in order to do both things in parallel?

api/src/routes/inference/expand-handler.ts
```diff
@@ -0,0 +1,258 @@
...
+
+/**
+ * Builds the AI context for the handler and middleware.
+ *
+ * @param {SourceFunctionResult | null} handler - The handler function result.
+ * @param {FindSourceFunctionsResult} middleware - The middleware functions result.
+ * @returns {Promise<[string | undefined, string | undefined] | [null, null]}
+ *          A promise that resolves to the handler and middleware context, or null if not found.
+ */
+async function buildAiContext(
```
flenter: I was looking at this and eventually thought: maybe this shouldn't be a function. It's job is calling two separate functions and returning those results in an array (maybe it would be better if it's an object). But looking at the code i was thinking, why not let other code call `buildHandlerContext(handler)` and `buildMiddlewareContext(middleware)` directly. 

Another reason why i was thinking it shouldn't be a function is that according to the jsdoc above, this function can return a array with either one of the two values being a string and the other undefined or both values are null. I am wondering why this is the case, perhaps and maybe the functions that are being called here can describe when they return a string/undefined or null.

api/src/routes/inference/expand-handler.ts
```diff
@@ -0,0 +1,258 @@
...
+  handler: SourceFunctionResult | null,
+  middleware: Array<SourceFunctionResult>,
+) {
+  const handlerContextPromise = handler
+    ? buildHandlerContext(handler)
+    : Promise.resolve(undefined);
+
+  const middlewareContextPromise = middleware
+    ? buildMiddlewareContext(middleware)
```
flenter: an array is true(thy), so this will always be executed

flenter (approved):
Nice!

api/package.json
actualwitch: typescript is already in root deps, might be that it would work without this

brettimus: indeed it is - but we need it as a prod dependency, and good to be explicit.

i'm rather fearful of how big our npx bundle has become

api/src/lib/ai/openai.ts
brettimus: i like to be able to quickly uncomment things to check outputs 😬

api/src/lib/expand-function/ast-helpers/ast-helpers.ts
brettimus: yeah gladly - i should probably just remove this "type" altogether since it isn't used how i intended to use it. will look

api/src/lib/expand-function/ast-helpers/ast-helpers.ts
brettimus: not really — the `ts.SourceFile` type is more specialized than a `ts.Node` as far as i understand. 

the sourceFile needs to be used to find the position of the node.

i suppose the function could accept an additional argument, `node`, then this could indeed be simplified.

api/src/routes/inference/expand-handler.ts
brettimus: nice yeah i removed. there were excessive functions in here due to some code shuffling when i realized i need to speed things up a bit. this function used to do more things, but then got gutted. its gonneeee

api/src/lib/expand-function/ast-helpers/ast-helpers.ts
flenter: Right, having the node option would probably mean that we'd need to describe why there's an optional node parameter so this wouldn't really improve things. I just don't like nested code/functions, but this is fine.

PR #297, opened 2024-10-02
Fix problem with enabling studio proxy (webhonc)

brettimus:
Fixes a few problems with enabling webhonc in the Studio. 

Currently, when you enable WebHonc, you get a `false` id back. This was because of a few problems:

- `POST /settings` was not accurately determining when the webhonc setting flag changed
- The UI did not properly invalidate the query for the webhonc id, in order to request the latest proxy id


actualwitch (approved):
🪿

PR #298, opened 2024-10-02
FP-4181: Change api port used by studio when switching workspaces

stephlow:
Includes: FP-4136


api/src/app.ts
```diff
@@ -21,6 +22,13 @@ export function createApp(
 ) {
   const app = new Hono<{ Bindings: Bindings; Variables: Variables }>();
 
+  app.use(
```
hatchan, line 25: Can you create a ticket to track that we make this a bit more restrictive

fpx/src/api.rs
```diff
@@ -74,6 +75,7 @@ impl Builder {
             )
             .with_state(api_state)
             .fallback(StatusCode::NOT_FOUND)
+            .layer(CorsLayer::new().allow_origin(Any))
```
hatchan: We should make this a opt-in, since it will be used by fpx-cli and fpx-workers as well. So just enable it for fpx-app for now.

hatchan (approved):

PR #300, opened 2024-10-04
Improve Route Updates

brettimus:
## Changes

- Increase delay before firing off probe
- Send latest routes every time a new request comes into the monitored Hono app
- Add unit tests (and vitest) to client library, for the route probe functionality
- Update CI to run client library unit tests

## To Try

- [x] Make sure the frontend websocket connection is still live when page is not in focus 

### Monitoring `.wrangler/tmp/dev-*`

okay so i want to make sure we're also monitoring this path in the watchDir:

`.wrangler/tmp/dev-*/index.js`

however, we need to only monitor the most recent dev-* folder in that directory...

we should modify the code in `api/src/lib/utils` to make sure we also respond to these changes ((and if that folder is present it's probably the only one we want to monitor right?))

### DIFF
--- a/packages/client-library-otel/src/instrumentation.ts
+++ b/packages/client-library-otel/src/instrumentation.ts
@@ -129,15 +129,17 @@ export function instrument(app: HonoLikeApp, config?: FpxConfigOptions) {
             return await originalFetch(request, rawEnv, executionContext);
           }
 
-          // If the request is from the route inspector, respond with the routes
+          // If the request is from the route inspector, send latest routes to the Studio API and respond with 200 OK
           if (isRouteInspectorRequest(request)) {
             logger.debug("Responding to route inspector request");
-            return respondWithRoutes(webStandardFetch, endpoint, app);
+            const response = await respondWithRoutes(
+              webStandardFetch,
+              endpoint,
+              app,
+              logger,
+            );
+            return response;
           }
-
-          // NOTE - We want to report the latest routes to FPX on every request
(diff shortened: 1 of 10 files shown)


packages/client-library-otel/src/instrumentation.ts
```diff
@@ -131,6 +135,10 @@ export function instrument(app: HonoLikeApp, config?: FpxConfigOptions) {
             return respondWithRoutes(webStandardFetch, endpoint, app);
           }
 
+          // NOTE - We want to report the latest routes to FPX on every request
+          //        so that we have an up-to-date list of routes in the UI
+          sendRoutes(webStandardFetch, endpoint, app);
```
flenter: I have a couple of ideas:
* maybe we can make it configurable whether this is executed?
* maybe this function should return a promise, so we can wait until the fetch is completed? Like using `waitUntil` or if that doesn't exist we could just add it to the `promsieStore`

brettimus: yeah i think pushing it into the promise store would be best solution, i'll add that

brettimus: @flenter i updated to use the promise store, and added some unit tests for the exports from `routes.ts` if you wanna take a look there

hatchan (approved):

flenter (approved):
LGTM, but can you also update the changelog?

stephlow (approved):

PR #304, opened 2024-10-05
Fix D1 auto instrumentation

brettimus:
A recent change in the D1 binding in miniflare requires a commensurate change in how we proxy the D1 binding. 

This change to the client library should fix issues that @Nlea was having getting D1 queries to show up in her fresh example app.


flenter (approved):
Don't forget to update the changelog

PR #305, opened 2024-10-07
Handle our workspace:* deps properly in static analysis code

brettimus:
Testing static analysis with goose-quotes in the example folder got tricky. We were trying to follow the defintion of `measure` back to the workspace dist (compiled js for the hono client lib), and that overloaded the language server.

This PR makes sure we ignore fpx workspace deps when doing static analysis 

***

Bonus: Do not follow the `arguments` identifier at all, as this can cause serious issues

### DIFF
--- a/api/src/lib/expand-function/ast-helpers/identifier-analyzer.ts
+++ b/api/src/lib/expand-function/ast-helpers/identifier-analyzer.ts
@@ -205,6 +205,11 @@ export function analyzeOutOfScopeIdentifiers(
       const parent = node.parent;
       const name = node.text;
 
+      // HACK - Just ignore the `arguments` identifier, it can cause us heartache following this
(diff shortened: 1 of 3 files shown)


flenter (approved):
LGTM, but i was thinking: wouldn't it be good idea to create issues for workspace package support (as well as support for parsing argument)?

oscarvz (approved):

brettimus:
> LGTM, but i was thinking: wouldn't it be good idea to create issues for workspace package support (as well as support for parsing argument)?

good point, i created an issue for handling workspace packages. however, for `arguments` i'm comfortable just ignoring that identifier for now — the only place this would break is if someone defined a variable called `arguments` in a higher level scope, and knowledge of that identifier's value would improve the request generation

PR #313, opened 2024-10-09
Add ollama support

laulauland:
A cursory test of the Llama 3.1 model shows that it has become quite capable of supporting some of our request generation use cases. This makes for a good case to add support for using the Fiberplane Studio with Ollama.

To test:

- Set up ollama locally, make sure you have llama 3.1 (8B) running: `ollama run llama3.1`
- Rebuild types (`pnpm build:types` from the root)
- Go to settings, and set your model provider to `ollama`


api/src/lib/ai/ollama.ts
```diff
@@ -0,0 +1,173 @@
...
+      {
+        role: "system",
+        content: getSystemPrompt(persona, "ollama"),
+      },
+      {
+        role: "user",
+        content: userPrompt,
+      },
+    ],
+    temperature: 0.04,
```
brettimus: I reduced temperature on the model for more predictable outputs, but this does mean it is more likely to keep returning the same request payload for a given route / handler

api/src/lib/ai/ollama.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  const makeRequestCall = message.tool_calls?.[0];
+  const toolArgs = makeRequestCall?.function?.arguments;
+
+  logger.debug("ollama toolArgs", JSON.stringify(toolArgs, null, 2));
+
+  // HACK - llama 3.1 (8B) might get confused and just respond with JSON describing the tool call.
+  //        So we should try to parse the content as a tool call
+  //        ((this does not really work as a fallback yet, but usually the json is malformed anyhow))
+  if (!toolArgs) {
```
brettimus: this doesn't really work as a fallback, might be worthwhile just to delete ¯\\_(ツ)_/¯ 

that said, it's worthwile to keep the warning log

api/src/lib/ai/ollama.ts
```diff
@@ -0,0 +1,173 @@
...
+      ? JSON.parse(message?.content ?? "")
+      : null;
+    logger.debug("ollama parsedContent", parsedContent);
+    // TODO - Validate the parsed content as a tool call, this object could take a couple of different shapes
+    return parsedContent;
+  }
+
+  try {
+    const parsedArgs = toolArgs ? JSON.parse(toolArgs) : null;
+    // HACK - llama 3.1 does not deal well with nested json objects, and will return each property as its stringified value.
```
brettimus: lots of hacks here to get the function call parameters into the right shape. 

the crux of it is: llama 3.1 8b is not good at respecting the json schema, so a lot of the times it will return nested json stringified params

we do a little hacky magic here to detect if a certain function parameter needs to have its value json parsed

api/src/lib/ai/ollama.ts
```diff
@@ -0,0 +1,173 @@
...
+      for (const key in parsedArgs) {
+        // We do not want to json parse the body field - it should be a string
+        if (key === "body") {
+          continue;
+        }
+
+        const value = parsedArgs[key];
+        if (typeof value === "string" && isJson(value)) {
+          const parsedValue = JSON.parse(value);
+          if (key === "headers" && Array.isArray(parsedValue)) {
```
brettimus: the model consistently has problems with adhering to our headers schema, as an improvement, we might need to give it a different schema, then parse the headers it returns into the format we typically expect

api/src/lib/ai/prompts.ts
```diff
@@ -302,6 +308,176 @@ Use the tool "make_request". Always respond in valid JSON.
...
+/**
+ * A QA (hostile) tester prompt.
+ *
+ * This prompt is used to generate requests for the API.
+ * It is a QA tester, who tries to break your api.
+ *
+ * NOTE - I had to stop instructing the AI to create very long data in this prompt.
+ *        It would end up repeating 9999999 ad infinitum and break JSON responses.
+ */
+export const LLAMA_3_8B_QA_PARAMETER_GENERATION_SYSTEM_PROMPT = cleanPrompt(`
```
brettimus, line 396: NOTE!!! I have not updated the "hostile" prompt at all yet, it may run into issues

api/src/lib/ai/prompts.ts
```diff
@@ -302,6 +308,176 @@ Use the tool "make_request". Always respond in valid JSON.
 ***Don't make your responses too long, otherwise we cannot parse your JSON response.***
 `);
 
+export const LLAMA_3_8B_FRIENDLY_PARAMETER_GENERATION_SYSTEM_PROMPT =
```
brettimus, line 310: this prompt has been modified from the OG system prompt we usually send to anthropic and openai

i found that reducing the scope of what we're trying to do worked better. so, we ONLY tell it to deal with json bodies, not worry about text + form data + etc

api/src/routes/inference/inference.ts
```diff
@@ -63,17 +61,22 @@ app.post(
...
     // Uncomment console.time to see how long this takes
     // It should be slow on the first request, but fast-ish on subsequent requests
     //
     // console.time("Handler and Middleware Expansion");
     const [handlerContextPerformant, middlewareContextPerformant] =
-      await expandHandler(handler, middleware ?? []).catch((error) => {
-        logger.error(`Error expanding handler and middleware: ${error}`);
-        return [null, null];
-      });
+      // HACK - Ditch the expand handler for ollama for now, it overwhelms llama 3.1-8b
```
brettimus, line 73: for the purposes of a demo, use a simplistic app and turn off static analysis.

keeps response times lower, and prevents the model from getting confused about tool calls (from my limited testing)

brettimus: however, it was able to handle smaller bits of additional context about handler functions, fwiw

packages/types/src/settings.ts
```diff
@@ -66,6 +68,8 @@ export const SettingsSchema = z.object({
       baseUrl: z.union([z.literal(""), z.string().trim().url()]).optional(),
     })
     .optional(),
+  ollamaModel: z.string().optional(),
+  ollamaBaseUrl: z.string().optional(),
```
brettimus: note: if you do `z.string().optional().default("http://")`, this gets converted into a type where the property is no longer considered optional, which breaks an assumption about the Settings type, i.e., that `Settings` is a partial object 

providing a default value marks the property as non-optional

studio/src/pages/SettingsPage/AISettingsForm.tsx
```diff
@@ -199,73 +210,79 @@ export function AISettingsForm({
...
-                                  <FormControl>
-                                    <Input
-                                      value={field.value ?? ""}
-                                      onChange={field.onChange}
-                                    />
-                                  </FormControl>
-                                </div>
-                              )}
-                            />
+                            {(providerField.value === "openai" ||
```
brettimus: hide API KEY and BASE URL for ollama

mies:
we should show a nice message when you've picked a model in ollama that has not been pulled

api/package.json
```diff
@@ -42,20 +42,16 @@
     "@ai-sdk/anthropic": "^0.0.51",
     "@ai-sdk/mistral": "^0.0.42",
     "@ai-sdk/openai": "^0.0.66",
+    "@anthropic-ai/sdk": "^0.24.3",
     "@clack/core": "^0.3.4",
     "@clack/prompts": "^0.7.0",
-    "@neondatabase/api-client": "^1.10.3",
-    "giget": "^1.2.3",
```
brettimus, line 48: do not remove any of the deps that the `cli` folder also has (giget, open, picocolors, oslo) - we need all of the cli deps copied here for  the create flow to work. it's hack, i know, and we should just move the cli folder into `api` if we wanna keep `create` here

brettimus: oops wait i see they were just reordered. closing

mies (approved):

PR #317, opened 2024-10-10
Ensure that the config has a default listen port

stephlow:
Fixes FP-4199

### DIFF
--- a/fpx-app/src/api_manager/legacy.rs
+++ b/fpx-app/src/api_manager/legacy.rs
@@ -13,13 +13,13 @@ pub struct ApiManager {
 }
 
 impl ApiManager {
-    /// Start a API server. If a API pid is already set, then that will first be
+    /// Start an API server. If an API pid is already set, then that will first be
     /// shutdown.
     pub fn start_api(&self, fpx_config: FpxConfig) {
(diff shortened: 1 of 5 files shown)


fpx-app/src/api_manager/legacy.rs
```diff
@@ -28,9 +28,9 @@ impl ApiManager {
 
         // Create some environment variables overrides based on the fpx.toml
         let mut envs: Vec<(&str, String)> = vec![];
-        if let Some(listen_port) = fpx_config.listen_port {
-            envs.push(("FPX_PORT", listen_port.to_string()));
-        }
+
+        let listen_port = fpx_config.listen_port();
+        envs.push(("FPX_PORT", listen_port.to_string()));
```
hatchan: We can move this initialization to line 30 already.

fpx/src/config.rs
```diff
@@ -48,6 +50,10 @@ impl FpxConfig {
 
         Ok((config, config_file_path))
     }
+
+    pub fn listen_port(&self) -> u32 {
```
hatchan, line 56: Can you add a little bit of documentation to the fn, related to what it does.

hatchan (approved):

PR #319, opened 2024-10-14
Fiberplane Auth integration

brettimus:
image

The basic flow/components are:

- Deploy auth service (`fp-services`), which:
  - Has credentials of GitHub OAuth application
  - Can create JWTs for users who log in with GitHub
  - Can forward JWTs to local studio
  - Has authenticated routes for AI request generation
  
- Configure Studio to run a tiny local auth server in the background, which gives us a predictable port to communicate with the auth service mentioned above
- Modify Studio UI and API to allow users to log in
- Modify Studio settings page for AI Request Generation to include details about using FP as the "model"
- Implement a basic profile page

## TODO
- [x] Implement AI calls agains FP service
- [x] Show remaining requests in UI
- [x] Propagate expiry


api/src/constants.ts
```diff
@@ -5,3 +5,11 @@ export const DEFAULT_DATABASE_URL = "file:fpx.db";
...
   process.env.FPX_WATCH_DIR ?? process.cwd(),
 );
+
+/** The port on which to run our ephemeral, local authentication server */
+export const FPX_AUTH_SERVER_PORT = 3579;
+
+export const FPX_PORT = +(process.env.FPX_PORT ?? 8788);
+
+export const FPX_AUTH_BASE_URL =
+  process.env.FPX_AUTH_BASE_URL || "http://localhost:3578";
```
brettimus: TODO: Change default

brettimus:
Worth noting: Auth does not work universally if you have multiple instances of Studio running — since it is using a predefined port.

stephlow (approved):

api/src/lib/fp-services/errors.ts
```diff
@@ -0,0 +1,38 @@
+// NOTE - Copy-pasted from fp-services/src/constants.ts
+export const ERROR_TYPE_UNAUTHORIZED = "UnauthorizedError";
+export const ERROR_TYPE_NOT_FOUND = "NotFoundError";
+export const ERROR_TYPE_INVALID_TOKEN = "InvalidTokenError";
+export const ERROR_TYPE_TOKEN_EXPIRED = "TokenExpiredError";
+
+// New error classes
```
brettimus:
```suggestion
```

PR #323, opened 2024-10-16
Use optional chains on app_responses because it might be falsy

brettimus:
The `app_responses` property on a proxied request might be blank if a request is cancelled or times out. 

Right now, if that's the case, the UI will break. This PR adds some defense agains that


flenter (approved):

oscarvz (approved):
The type in `ProxiedRequestResponse` should probably be updated to reflect it can turn out to be `undefined`. Probably out of scope of this PR, but I'd be happy to help out

brettimus:
> The type in `ProxiedRequestResponse` should probably be updated to reflect it can turn out to be `undefined`. Probably out of scope of this PR, but I'd be happy to help out

true, we should probably decode that response properly. created issue: https://linear.app/fiberplane/issue/FP-4210/update-proxedrequestresponse-type

PR #345, opened 2024-10-29
Add alternative analysis logic

flenter:
This pr does the following:

* removed the requirement for manually building the types package (several of the packages get build after running `pnpm install`)
* changed the way api logs (visually logs) to the console:
image

* disallow the usage of console.log in the api codebase  (`logger.[method]` should be used)
* added a `@fiberplane/source-analysis` package. This is where the bulk of action is happening
  * it allows the new logger to be shared/used. The api passes a logger to this package which includes a tag so all logs related to the source-analysis package are tagged.


api/src/routes/inference/inference.ts
```diff
@@ -141,22 +141,29 @@ app.post(
...
-        // TODO handle openApiSpec
-        openApiSpec: openApiSpec ?? undefined,
-        // middleware: middleware ?? undefined,
-        middleware: undefined,
-        middlewareContext: middlewareContextPerformant ?? undefined,
-      });
+      await generateRequestWithAiProvider(generateRequestOptions);
+
+    // Log the request data for future inspection
+    await db.insert(schema.aiRequestLogs).values({
```
brettimus, line 102: NOTE - probably only want to run this in dev

packages/source-analysis/src/RoutesResult/generate.ts
```diff
@@ -291,7 +291,7 @@ function extractRouteTreeContent(includeIds: boolean, resource: RouteTree) {
 
 function serializeFileMap(fileInfo: FileInfo): string {
   const imports = Array.from(
-    fileInfo.modules.entries().map(([importPath, modules]) => {
+    Array.from(fileInfo.modules.entries()).map(([importPath, modules]) => {
```
brettimus: this fixed the error i was seeing:

```
Error expanding handler and middleware: TypeError: fileInfo.modules.entries(...).map is not a function
```

needed to make sure `entries` was an array

flenter: I added the `Array.from` there for that reason. I'm not quite sure why i never ran into any issue with this. But I think we can remove one Array.from call

brettimus (approved):

stephlow (approved):

PR #364, opened 2024-11-18
Fix shared types package

laulauland:
Shared types have been a bit of a pain to work with. This is an attempt at fixing that.

The gist of it is that we are doing some package.json voodoo to point to `src/*` directly as opposed to `dist/*`. TypeScript imports then resolve in the respective consumers while developing.

Because we are publishing the package to `npm`, we change the `exports` field in `package.json` to point to the `dist/*` files in the `publishConfig` section.

### DIFF
--- a/api/bin/cli.js
+++ b/api/bin/cli.js
@@ -23,7 +23,7 @@ const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
 const args = process.argv.slice(2);
-// Handle emptystring script (I don't think this happens in practice but let's be defensive)
+// Handle empty string script (I don't think this happens in practice but let's be defensive)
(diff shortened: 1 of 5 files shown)


brettimus:
so it seems like this fixes the issue of not having built types locally, but we could still have the issue where we accidentally point to an unpublished version of the types when we publish studio?

brettimus (approved):
worked for me locally, when deleting the `dist` from the types package and then spinning up the api

laulauland:
Reverted the same change for the source analysis package for now as we need it to follow the same conventions of the `api` to make it work (using `*.js` postfixes for imports). Have proven this works though so might wanna ship it in the next PR. cc @flenter @brettimus

PR #369, opened 2024-11-19
Detect openapi definitions from hono-zod-openapi apps

brettimus:
> **UPDATE TYPES PACKAGE AFTER MERGE**
> **UPDATE HONO-OTEL AFTER MERGE**

## Description

Automatically report openapi spec from the client library to the studio api.

Whenever a route probe response is received, Studio will try to match any incoming routes with a corresponding openapi definition for that route.

If one is found, a json object describing the parameters and response for the route is then saved in the database on a special column for the route. This mini-spec can be used to generate request parameters, which is useful, since the source code analysis can't do much to help with OpenApiHono apps (more on that below).

The spec for a route, if present, also will render a Docs tab for that route.

### Feature Flagged

Adds a section in the settings page where you can point to a route that hosts a json openapi definition.
This is a more flexible approach than auto-detecting the spec from the Hono-Zod-OpenAPI integration.

## Fun things along the way

- `OpenApiHono` is not picked up by source code analysis (makes sense, since it is not of type `Hono`)
- Had to add a special header that allows the client library to skip instrumentation for an incoming request. Otherwise we'd enter an infinite loop when explicitly looking up a spec from the target api ((_note that the "custom spec url" setting  is behind a feature flag for now))
- Because of said infinite loop, had to finally add a limit to the number of traces we return from the api. (My infinite loop produced close to 90K new spans, which ground the UI to a halt
- Basic auth is wonky in studio 

## Considerations

**Manually pulling the spec versus automatically receiving it** — My first implementation relied on the user explicitly telling us where their spec was located, then we would try to pull it in to find matching specs for routes. However, things are a lot smoother if we only support apps built with `HonoOpenAPI` for now. In that case, we can ignore the URL specified on the settings page. Also, if we want to support "give us a URL and we'll try syncing", then we need to be mindful of things like the `ALL` method. It's a lot more work to support a link to a spec instead of relying on the Hono OpenAPI integration to report it to us.

**Autopopulated parameters** — This is tricky given that when you click on a Route, we might end up redirecting you to the last-made-request for that route. In which case the expected behavior is not clear. Should we merge expected query parameters from the OpenAPI spec with the _actual_ query parameters made in the last request? 

## Screenshots

image

image

image

image

## TODO
- [ ] Test headers
- [x] Show documentation for route somewhere

- [ ] ~Handle `.all` method~
- [ ] ~Only allow openapi integration for versions of client library past 0.4.1~

- [x] Expand schema definitions when possible into openApiSpec
- [x] Explain OpenAPI integration on the settings page
- [x] Auto-enable openapi integration when we can


api/src/lib/openapi/fetch.ts
```diff
@@ -0,0 +1,60 @@
...
+  const resolvedSpecUrl = resolveSpecUrl(specUrl);
+  if (!resolvedSpecUrl) {
+    logger.debug("No resolved OpenAPI spec URL found");
+    return null;
+  }
+  logger.debug(`Fetching OpenAPI spec from ${resolvedSpecUrl}`);
+  const response = await fetch(resolvedSpecUrl, {
+    headers: {
+      // NOTE - This is to avoid infinite loops when the OpenAPI spec is fetched from Studio
+      //        We need to make sure that the user has instrumented their app with @fiberplane/hono-otel >= 0.4.1
```
brettimus: this part is really crucial - we need to validate that the user has the correct version of hono-otel

api/src/routes/traces.ts
```diff
@@ -35,6 +35,7 @@ app.get("/v1/traces", async (ctx) => {
   const spans = await db.query.otelSpans.findMany({
     where: sql`inner->>'scope_name' = 'fpx-tracer'`,
     orderBy: desc(sql`inner->>'end_time'`),
+    limit: 1000,
```
brettimus, line 38: this is a little left field but i finally hit the issue where i had so many spans it effectively broke the UI. i put a limit of 1000 here which seems okay but... who knows

examples/openapi-zod/src/db/schema.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const users = sqliteTable("users", {
+  id: integer("id", { mode: "number" }).primaryKey(),
+  name: text("name").notNull(),
+  age: integer("age").notNull(),
+  createdAt: text("created_at").notNull().default(sql`(CURRENT_TIMESTAMP)`),
+  updatedAt: text("updated_at").notNull().default(sql`(CURRENT_TIMESTAMP)`),
+});
+
+// TODO - Figure out how to use drizzle with "@hono/zod-openapi"
```
brettimus, line 13: worth noting: i wanted to use `drizzle-zod` to share schemas between the database and the api, in order to reuse a database schema as, e.g., the payload for a route

however, the openapi integration requires that you use a special version of zod that they export, which provides some additional fields that are necessary for the schema to be converted to an openapi schema

this forces you to redefine (as far as i can tell) the schemata, which is kind of good practice to separate api schemas from database schemas, but was a little bit of a pain for me considering i didn't care about that separation at this step

.gitignore
```diff
@@ -4,6 +4,9 @@ shared/dist
 .envrc
 .cursorrules
 
+# Cursor
+.cursorrules
```
flenter, line 7: This line already exists in this file

api/src/lib/openapi/openapi.ts
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Enriches API routes with their corresponding OpenAPI specifications by fetching OpenAPI definitions
+ * and mapping the routes in the spec to Hono routes.
+ *
+ * If an `openApiSpec` is provided, it will be used instead of fetching the latest spec from the database.
+ *
+ * This function handles both single routes and arrays of routes.
+ *
+ * @param db - LibSQL database instance containing the OpenAPI specifications. Used to fetch the latest spec.
+ * @param routes - Single route or array of routes to be enriched. Each route should contain path, method,
```
flenter: Maybe i'm reading this text the wrong way, but  the way I understand this doc is as follows:
```
route: Route | Array<Route>
```
However it's always an array (though maybe with a singular route)

And that makes me think, I've seen in other prs, that the JSDocs are out of date even in the PR's that introduce them. Maybe it's better to leave the types out for parameters/return values?  I do like the part of the JSDoc where you describe what a function is intended for but types/parameters/return values  to me seems like we need to maintain an additional thing (like for instance if you refactor/rename a parameter you'd need to manually also update the docs).

examples/openapi-zod/src/index.ts
```diff
@@ -0,0 +1,266 @@
...
+  DB: D1Database;
+};
+
+const app = new OpenAPIHono<{ Bindings: Bindings }>();
+
+// Schema that defines presence of an ID in the path
+const UserIdPathParamSchema = z.object({
+  id: z
+    // Path params are always strings
+    .string()
```
flenter, line 22: You can also do:
`z.number({ coerce: true})` and the id will be converted to a number

studio/src/pages/RequestorPage/CommandBar/CommandBar.tsx
```diff
@@ -0,0 +1,143 @@
...
+  DialogTitle,
+} from "@/components/ui/dialog";
+import { Icon } from "@iconify/react/dist/iconify.js";
+import React from "react";
+import { useRequestorStore } from "../store";
+
+type CommandBarProps = {
+  open: boolean;
+  setOpen: (open: boolean) => void;
+  // onGenerateRequest?: (prompt?: string) => void;
```
flenter: Can go?

studio/src/pages/RequestorPage/RequestPanel/RouteDocumentation/RouteDocumentation.tsx
```diff
@@ -0,0 +1,301 @@
...
+  openApiSpec,
+}: RouteDocumentationProps) {
+  const { parameters, requestBody, responses, description, summary } =
+    openApiSpec;
+
+  console.log("openApiSpec", openApiSpec);
+  return (
+    <ScrollArea className="h-full pb-8">
+      <div className="p-2">
+        {/* Description Section - Updated styling */}
```
flenter: Is this still relevant?  I think it can go, not?

flenter (approved):
lgtm

examples/openapi-zod/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "name": "zod-openapi",
+  "scripts": {
+    "dev": "wrangler dev src/index.ts",
+    "deploy": "wrangler deploy --minify src/index.ts",
+    "db:generate": "drizzle-kit generate",
+    "db:migrate": "wrangler d1 migrations apply zod-openapi --local",
+    "db:studio": "drizzle-kit studio",
+    "db:touch": "wrangler d1 execute zod-openapi --local --command='SELECT 1'",
+    "db:migrate:prod": "GOOSIFY_ENV=production drizzle-kit migrate",
+    "db:studio:prod": "GOOSIFY_ENV=production drizzle-kit studio"
```
flenter: I would like it if we'd had some type checking/linting enabled also for our example projects.

examples/openapi-zod/src/index.ts
brettimus: wasn't sure how that'd play with validation — worth toying around with it

studio/src/pages/RequestorPage/RequestPanel/RouteDocumentation/RouteDocumentation.tsx
brettimus: indeed - trashed all the comments

examples/openapi-zod/package.json
brettimus: we could add it to the main biome file but seems out of scope for this pr!

api/src/lib/openapi/openapi.ts
brettimus: true true. i'll avoid parameter docs in the future, because this was indeed a crufty comment

PR #374, opened 2024-11-26
Add instrumentation to RPC calls to other workers

brettimus:
## Overview

This PR focused on getting the client library to record spans for any (sub)workers that were invoked via the cloudflare service binding interface (https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/)

UI changes will happen in a follow-on PR

image

## Running the test app locally

```sh
# in one terminal
cd examples/service-bindings/meow
cp .dev.vars.example .dev.vars
pnpm dev

# in another terminal
cd examples/service-bindings/woof
pnpm dev

# connect your local studio api:
cd api
FPX_WATCH_DIR=../examples/service-bindings/meow pnpm dev
```

## Changes

- Add a sample app that uses worker service bindings (https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/)
- Detect service-bound workers when we patch Cloudflare bindings and instrument rpc calls
- Remove `swc` from the client-library build step since its compilation target broke our proxy objects for workers
- Client library should treat any `thennable` as a Promise, instead of checking for `instanceof Promise`

## TODO

- [x] Test a production build of the client library (make sure removing swc didn't break anything)
- [ ] Test with `fetch`
- [ ] Test with bound durable objects
- [x] Resolve issue where span length is always 0


laulauland (approved):

laulauland:
Tested with the provided example - no issues spotted 👍

One thing, if my bound worker makes an outgoing fetch call, those calls are not intercepted. I understand this might be out of scope for this PR but just making a note.
image

image

Already approved, just gonna test quickly with a DO example.

laulauland:
Doesn't work with Durable Objects bindings currently, but that can be a separate PR too

flenter (approved):

PR #375, opened 2024-11-26
App routes file source tree using code analysis

stephlow:
~Index based keyboard navigation is broken atm in the file tree~

- Enable analysis by default
- Tabs to switch between tree and list view
- Searchable
- Fixed refresh issue with query key's
- Collapses tree nodes if there's only one child 

Expanded example:

https://github.com/user-attachments/assets/031c6a36-f769-43b1-8cc0-b3cb5de6589e

Single file example:

https://github.com/user-attachments/assets/221bfeae-4dd9-4dba-b00b-69afe5a5ca0a


flenter:
I've tweaked the UI a bit in order to improve the looks:
image

mies:
Looks great! How difficult would it be to fix the keyboard navigation?

flenter:
I have fixed the issue where the currently active route is highlighted (though not in a pretty way). And I'm now looking at the selection. It is a bit more involved because keyboard navigation also can also go to the custom routes

mies:
sounds good! Keyboard Navigation would still be a nice to have for today's release 🚀

flenter:
I've got a basic version of it working. However when switching between different views, all collapsed/expanded state gets reset and also: keyboard navigation does not take into consideration that you can collapse tree items. So it could benefit from a refactor

studio/src/pages/RequestorPage/types.ts
```diff
@@ -47,7 +47,7 @@ export const ProbedRouteSchema = z.object({
   currentlyRegistered: z.boolean(),
   registrationOrder: z.number().default(-1),
   routeOrigin: z.enum(["discovered", "custom", "open_api"]),
-  openApiSpec: z.string().optional(),
+  openApiSpec: z.string().nullish().optional(),
```
flenter, line 51: This the second definition of the `app route` we should investigate if we can delete one of them. But given the deadline, i haven't done this for this PR

flenter:
Keyboard navigation should be fixed now

laulauland (approved):

PR #377, opened 2024-11-27
Enable observability on the webhonc

brettimus:
This is to make our local version match the deployed version config, as I updated webhonc to have workers logs via the CF dashboard


webhonc/wrangler.toml
```diff
@@ -1,7 +1,9 @@
-#:schema node_modules/wrangler/config-schema.json
```
brettimus, line 1: had to remove this — it doesn't contain `observability` so will complain

flenter (approved):

oscarvz (approved):

PR #378, opened 2024-11-27
Dependency updates and other cleanup

laulauland:
An in-progress PR tracking the sweeping dependency bump across our TS packages.

TODO:
- [x] fix up type and lint issues


hatchan (approved):

fp-services/package.json
```diff
@@ -22,23 +22,23 @@
...
+    "hono": "^4.6.12",
+    "jose": "^5.9.6",
+    "tsx": "^4.19.2",
+    "vitest": "^2.1.6",
     "zod": "^3.23.8"
   },
   "devDependencies": {
-    "@biomejs/biome": "^1.8.3",
-    "@cloudflare/workers-types": "^4.20241004.0",
+    "@biomejs/biome": "^1.9.4",
```
oscarvz, line 39: Biome packages should be locked to a single version, and the config `biome.jsonc` should be updated to make use of the updated schema.

oscarvz (approved):
Apart from one comment, lgtm 🚀

fp-services/package.json
laulauland: updated! All point to the same at the root of the monorepo now

PR #385, opened 2024-12-03
Fix invocations of `fetch` with Worker service bindings

brettimus:
Updates the client library to correctly proxy `fetch` calls to service bindings

The issue was that calling `fetch` on a worker service binding required the proxied `fetch` to still be bound to the WorkerEntrypoint. Basically, the client library needs to have some special logic for when it wraps `fetch` in a call to `measure`

### DIFF
--- a/examples/service-bindings/woof/src/index.ts
+++ b/examples/service-bindings/woof/src/index.ts
@@ -17,8 +17,8 @@ export class WoofWorker extends WorkerEntrypoint {
 
   // From Cloudflare docs: "Currently, entrypoints without a named handler are not supported"
   // TODO - Check if this is still the case that we need a fetch handler?
-  async fetch() {
-    return new Response(null, { status: 404 });
+  async fetch(request: Request | string) {
+    return fetch("https://placegoose.mies.workers.dev/geese");
   }
 }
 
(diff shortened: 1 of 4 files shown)


mies (approved):

brettimus:
waiting for nele to test with her repo before merging and cutting new release

examples/service-bindings/woof/src/index.ts
```diff
@@ -17,8 +17,8 @@ export class WoofWorker extends WorkerEntrypoint {
 
   // From Cloudflare docs: "Currently, entrypoints without a named handler are not supported"
   // TODO - Check if this is still the case that we need a fetch handler?
-  async fetch() {
-    return new Response(null, { status: 404 });
+  async fetch(request: Request | string) {
```
flenter: I guess there's no linter checking this code, but if the parameter type is something important, maybe rename `request` to `_request`?

flenter (approved):

examples/service-bindings/woof/src/index.ts
brettimus: yeah so you have to pass a string or Request otherwise the fetch call will fail. i will put an underscore!

brettimus:
```suggestion
  async fetch(_request: Request | string) {
```

PR #388, opened 2024-12-05
Add tooltip for sidebar toggle

brettimus:
We didn't have a tooltip explaining the keyboard shortcut for the sidepanel (thanks @ambergristle)

image


laulauland (approved):
nice one

PR #399, opened 2024-12-09
Fix draft properties in blog

laulauland:
- **chore: explicitly set packagemanager**
- **add property draft so that only non-drafts get picked up during build**


brettimus:
could you run `format` in the `www` folder?

brettimus (approved):

flenter (approved):

PR #400, opened 2024-12-09
Various Fixes: Websocket heartbeat, historical request sorting, et al

brettimus:
## TODOs

- [ ] ~Refactor `loadHistoricalRequest`~
- [x] Document telegram bot example
- [x] Fix ordering of history
- [x] Update changelog
- [x] Update package.json

## Description

This PR does a few things:

- Adds proper `pong` responses for the heartbeat coming from the Studio frontend
- Fix an edge case where request bodies were not loaded from history into the UI
- Fix sorting of historical requests (partially)
- Fix default timestamps in our sqlite database to be iso-8601
- Adds a minimal telegram bot example to the monorepo

## Heartbeat

Our `pong` responses to `ping` messages were never being sent down the wire, which caused the `useRealtimeService` hook to think that the websocket connection was no longer open after a minute.

Changing the studio api's websocket server to treat `ping`s as regular messages resolved the issue.

I kept the old code as a comment with an explainer that it was not working properly.

## Load request bodies from historical requests

When poking around with the new example api, I hit a case where `loadHistoricalRequest` was not setting the body for certain requests from the history. I had to make an ugly function uglier, but resolve this issue by copy-pasting code ((working on a refactor now))

> **NOTE** This will still be broken when we have a trace but the request was not proxied through Studio API (via the UI or Webhonc)

## Fix Default Studio Api Database Timestamps

`CURRENT_TIMESTAMP` in sqlite was not iso8601. Changed the db schema default values to fix that, since sometimes we were adding iso timestamps manually to certain records, and then having a mix of timestamps messed up sorting in the UI

## New Example: Telegram bot

**Why?**

I wanted a webhook related example in our monorepo to test possible connection interruptions to webhonc.


studio/src/pages/RequestorPage/useRequestorHistory.ts
```diff
@@ -194,6 +194,19 @@ export function useRequestorHistory() {
             })),
           ),
         );
+
```
brettimus, line 197: this function badly needs a refactor, but i have other things to do this week 😬

mies (approved):

studio/src/pages/RequestorPage/utils.ts
```diff
@@ -16,17 +16,55 @@ export function sortProxiedRequestResponsesDescending(
...
+    return timestamp;
+  }
+
+  // If it matches SQLite format, convert it
+  if (sqliteRegex.test(timestamp)) {
+    return `${timestamp.replace(" ", "T")}Z`;
+  }
+
+  // Return original timestamp if format is unknown
+  return timestamp;
```
flenter, line 69: If we don't throw an error here, i think it might be good to log a warning here. So if developers have the console open, it does show that something unexpected happened .

flenter (approved):

PR #410, opened 2024-12-12
Adds support for collections

flenter:
Adds in:
* collections
* confirmation dialogs
* shake animation
* updated `@fiberplane/fpx-types`

and so much more.  See it in action:

https://github.com/user-attachments/assets/9ee0eb7b-8153-45c5-bdf9-e58f903df4f6


api/drizzle/0022_confused_maelstrom.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE `collection_items` (
```
brettimus, line 1: huh - so this migration is repeated but it didn't cause any issues for me locally? 🤔

studio/tsconfig.json
```diff
@@ -28,6 +28,6 @@
       "@/*": ["./src/*"]
     }
   },
-  "include": ["src", "../client-library-otel/src/utils/wrapper.ts"],
+  "include": ["src", "../client-library-otel/src/utils/wrapper.ts", "src/pages/RequestorPage/CollectionForm.tsx"],
```
brettimus: can revert this? 
```suggestion
  "include": ["src", "../client-library-otel/src/utils/wrapper.ts"],
```

studio/tailwind.config.js
```diff
@@ -71,12 +73,27 @@ module.exports = {
           from: { height: "var(--radix-accordion-content-height)" },
           to: { height: "0" },
         },
+        shake: {
```
brettimus, line 76: the shake is back!

studio/src/queries/collections.ts
```diff
@@ -0,0 +1,316 @@
...
+    return CollectionItemSchema.parse(data);
+  });
+}
+
+function getErrorMessage(error: unknown): string {
+  if (typeof error === "string") {
+    return error;
+  }
+
+  if (
```
brettimus: could use `objectWithKey` for part of this check

studio/src/queries/collections.ts
```diff
@@ -0,0 +1,316 @@
...
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      appRouteId: routeId,
+      ...extraParams,
+    }),
+  }).then(async (r) => {
+    if (!r.ok) {
+      console.log("r", Array.from(r.headers.entries()));
```
brettimus:
```suggestion
```

(remove log)

studio/src/queries/collections.ts
```diff
@@ -0,0 +1,316 @@
...
+          const result = await (r.headers
+            .get("Content-Type")
+            ?.startsWith("application/json")
+            ? r.json()
+            : r.text());
+
+          throw resultToError(result);
+        }
+
+        await r.text();
```
brettimus: any reason to parse the body here? seems like we could just return

studio/src/queries/collections.ts
```diff
@@ -0,0 +1,316 @@
...
+        return undefined;
+        // return CollectionSchema.parse(json);
```
brettimus:
```suggestion
```

studio/src/queries/app-routes.ts
```diff
@@ -1,4 +1,4 @@
-import { useRequestorStore } from "@/pages/RequestorPage/store";
+import { useStudioStore } from "@/pages/RequestorPage/store";
```
brettimus, line 1: good rename

studio/src/pages/RequestorPage/store/utils.ts
```diff
@@ -6,6 +6,7 @@ import type { RequestorState } from "./types";
 export const _getActiveRoute = (state: RequestorState): ProbedRoute => {
   return (
     state.activeRoute ?? {
+      id: -1,
```
brettimus: or: `Number.NEGATIVE_INFINITY` 😄

studio/src/pages/RequestorPage/store/slices/types.ts
```diff
@@ -39,29 +41,39 @@ export interface RequestResponseSlice {
...
+      | "requestQueryParams"
+      | "requestRoute"
+      | "requestUrl"
+    >,
+  ) => void;
 }
 
 export interface RoutesSlice {
-  routes: ProbedRoute[];
+  appRoutes: ProbedRoute[];
```
brettimus, line 75: good rename

studio/src/pages/RequestorPage/store/slices/index.ts
```diff
@@ -2,6 +2,7 @@ export { aiSlice } from "./aiSlice";
 export { requestResponseSlice } from "./requestResponseSlice";
 export { routesSlice } from "./routesSlice";
 export { tabsSlice } from "./tabsSlice";
-export type { Store } from "./types";
+export type { StudioState as StudioStore } from "./types";
```
brettimus: StudioState or StudioStore?

api/src/db/schema.ts
```diff
@@ -175,3 +175,61 @@ export const aiRequestLogs = sqliteTable("ai_request_logs", {
   log: text("log", { mode: "json" }).notNull(),
   createdAt: timestamps.createdAt,
 });
+
+export const collections = sqliteTable("collections", {
+  id: integer("id", { mode: "number" }).primaryKey({ autoIncrement: true }),
+  name: text("name").notNull().unique(),
+  createdAt: text("created_at").notNull().default(sql`(CURRENT_TIMESTAMP)`),
+  updatedAt: text("updated_at").notNull().default(sql`(CURRENT_TIMESTAMP)`),
```
brettimus: would suggest using the "timetamps" helper since it'll be iso8601

```suggestion
  ...timestamps,
```

studio/src/App.tsx
```diff
@@ -23,22 +18,15 @@ export function App() {
...
             <Routes>
-              <Route path={REQUESTS_ROUTE} element={<RequestsPage />} />
               <Route
-                path={REQUEST_DETAILS_OTEL_ROUTE}
-                element={<RequestDetailsPage />}
+                path={INTERNAL_AI_LOGS_ROUTE}
+                element={<AiRequestLogsPage />}
               />
               <Route
-                path={REQUEST_DETAILS_TRACE_ROUTE}
```
brettimus, line 32: bye bye request details page

studio/src/Layout/Settings/SettingsScreen.tsx
```diff
@@ -1,6 +1,6 @@
 import { Button } from "@/components/ui/button";
 import { SettingsPage } from "@/pages/SettingsPage";
-import { Icon } from "@iconify/react/dist/iconify.js";
+import { Icon } from "@iconify/react";
```
brettimus, line 3: was there a reason we were importing from `dist/iconify.js`?

studio/src/hooks/useStickyLoading.ts
```diff
@@ -0,0 +1,21 @@
...
+      setSticky(true);
+      return;
+    }
+    const timeout = setTimeout(() => {
+      setSticky(false);
+    }, duration);
+    return () => clearTimeout(timeout);
+  }, [loading, duration]);
+
+  return sticky;
```
brettimus: should this return `sticky || loading`? 

if i understand correctly, it is a minimum duration for loading, but if `loading` is still true after `duration`, then it should still return `true`

brettimus (commented):
first round of comments -

studio/src/Layout/Settings/SettingsScreen.tsx
flenter: Not sure, it's vscode that imports it like that, but to me this seems brittle

api/drizzle/0022_confused_maelstrom.sql
flenter: ah right, i don't think this isn't used,. I will remove  the other one (19)

packages/types/src/api.ts
```diff
@@ -11,3 +11,40 @@ export const TraceListResponseSchema = z.array(TraceSummarySchema);
 export type TraceListResponse = z.infer<typeof TraceListResponseSchema>;
 
 export type TraceDetailSpansResponse = Array<OtelSpan>;
+
+export const CollectionSchema = z.object({
```
brettimus, line 15: Important! Gotta bump the version in `packages/types/package.json`

then edit the pull request description to remind us to publish the new version of fpx-types

brettimus:
Reminders:

- Update changelog
- Update api/package.json version
- Update packages/types/package.json version

Must remember to publish packages/types before releasing our next canary of studio

studio/src/pages/RequestorPage/store/slices/index.ts
flenter: Renaming in vscode has its limitations :) ... I noticed that in slices i also exported it differently. The idea is to call it call it state  and we keep it in a store

studio/src/queries/collections.ts
flenter: oh nice one.

studio/src/hooks/useInputFocusDetection.ts
```diff
@@ -19,11 +16,12 @@ export function useInputFocusDetection() {
       }
     };
 
-    document.addEventListener("focus", handleFocus, { capture: true, signal });
-    document.addEventListener("blur", handleBlur, { capture: true, signal });
+    document.addEventListener("focus", handleFocus, { capture: true });
+    document.addEventListener("focus", handleFocus, { capture: true });
```
brettimus: looks like the focus handler is registered twice here instead of also registering the `blur` handler

brettimus:
~bug: clicking on the "filter routes" text input does not activate focus on the component for me anymore ((haven't tried to figure out the cause yet))~

image


~keyboard shortcut still works for me~

***

fixed, needed to pass through click events from the placeholder elements

studio/src/pages/RequestorPage/AddRouteToCollection/AddToRouteForm.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+type Props = {
+  onSuccess: () => void;
+};
+
+export function AddToRouteForm(props: Props) {
+  const { data: collections, isLoading } = useCollections();
+  const { activeRoute } = useStudioStore("activeRoute");
+
+  // No active route? Then there's nothing to do here
```
brettimus, line 13: could be good to at least put a message in the UI here — otherwise if the user has typed something that doesn't match the routes, then all they see is an empty popover 

image

studio/src/pages/RequestorPage/AddRouteToCollection/AddToRouteForm.tsx
```diff
@@ -0,0 +1,49 @@
...
+    console.warn("No active route");
+    return null;
+  }
+
+  if (isLoading || !collections) {
+    return <div>Loading...</div>;
+  }
+
+  if (collections.length === 0) {
+    return <div>Empty</div>;
```
brettimus: probably for future, but maybe we always initialize a "starter" collection of some sort? 

maybe for now could just put some helper text like "Create a collection to get started" or something

studio/src/pages/RequestorPage/RequestPanel/RequestPanel.tsx
```diff
@@ -223,13 +251,18 @@ export const RequestPanel = memo(function RequestPanel(
...
+            updateCollectionItem({
+              requestHeaders: createObjectFromKeyValueParameters([]),
+            });
           }}
         />
         <KeyValueForm
           keyPlaceholder="header-name"
           keyValueParameters={requestHeaders}
           onChange={(headers) => {
-            setRequestHeaders(headers);
```
brettimus, line 232: 👀

brettimus:
bug: keyboard shortcuts for opening the logs and timeline panel are not working for me

studio/src/pages/RequestorPage/RequestPanel/RequestPanel.tsx
brettimus: i think we need this line

studio/src/utils/index.ts
```diff
@@ -216,44 +224,79 @@ export function isSensitiveEnvVar(key: string) {
...
+    return {
+      id: createParameterId(),
+      key,
+      value,
+      enabled: true,
+    };
+  });
+}
+
+export function createObjectFromKeyValueParameters<
```
brettimus, line 275: as far as i can tell, this is only used in cases where we sync with a collection - however, one thing to be wary of is that this function ignores whether a parameter is enabled/disabled. which is to say, it will copy/sync parameters that have been disabled by the end user

studio/src/queries/collections.ts
```diff
@@ -0,0 +1,287 @@
...
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: [COLLECTIONS_KEY] });
+    },
+    mutationKey: [COLLECTIONS_KEY, "update-collection-item"],
+  });
+
+  return mutation;
+}
+
+async function responseToJsonOrText(r: Response) {
```
brettimus, line 281: added this to factor out the ternary that was happening in most `if (!res.ok) { ... }` blocks

studio/src/pages/RequestorPage/store/utils.ts
```diff
@@ -6,6 +6,7 @@ import type { RequestorState } from "./types";
 export const _getActiveRoute = (state: RequestorState): ProbedRoute => {
   return (
     state.activeRoute ?? {
+      id: Number.NEGATIVE_INFINITY,
```
brettimus, line 9: hahaha i was being cheeky when i suggested `NEGATIVE_INFINITY` but i like that it's here

studio/src/pages/RequestorPage/CollectionSection/CollectionItemListItem.tsx
```diff
@@ -0,0 +1,160 @@
...
+            <span>Move to top</span>
+            <Icon icon="lucide:arrow-up-to-line" className="h-4 w-4" />
+          </CustomDropdownItem>
+          <CustomDropdownItem
+            onSelect={() => {
+              updateItem({
+                collectionId,
+                itemId,
+                extraParams: {
+                  position: Number.MAX_SAFE_INTEGER,
```
brettimus, line 134: this causes problems 😅  - it will always be at the bottom, even when i try to reorder something below it

brettimus: okay, issue was on the api side! fixed

studio/src/pages/RequestorPage/RequestorPageContent/RequestorPageContent.tsx
```diff
@@ -32,12 +42,93 @@ type RequestorPageContentProps = {
...
+  }, [itemId, collectionItemRef, appRouteRef, setRequestParams, updateMethod]);
+
+  useEffect(() => {
+    if (!item) {
+      return;
+    }
+
+    setRequestParams(item.app_requests);
+  }, [item, setRequestParams]);
+
```
brettimus: something is a little wonky with the introduction of these effects — i'm getting some flash-in behavior, especially where the UI will temporarily show the "activeResponse", then flash to the trace response. 

aside from that, the headers will temporarily flash to the list of headers made in the proxied request, then flash back to the og headers in the form. 

will try to resolve today

brettimus: at any rate, the effects could probably use comments that explain their purpose

brettimus: removing the second effect in this highlighted code did the trick for me. i'm not sure if it served a purpose, trying to look for things i might have broken

brettimus (approved):

PR #459, opened 2025-01-28
Update dependencies and quality of life improvement

hatchan:
Update a bunch of dependencies, get rid of some duplicate dependencies and ensure that the fpx binary is compiled as `fpx`, not `fpx-cli`.

### DIFF
--- a/fpx-cli/Cargo.toml
+++ b/fpx-cli/Cargo.toml
@@ -6,6 +6,10 @@ authors = { workspace = true }
 license = { workspace = true }
 repository = { workspace = true }
 
+[[bin]]
+name = "fpx"
+path = "src/main.rs"
+
 [features]
 embed-studio = [] # When enabled it will embed Studio from frontend/dist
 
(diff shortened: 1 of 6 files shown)


fpx-cli/Cargo.toml
```diff
@@ -58,7 +62,7 @@ strum = { version = "0.26", features = ["derive"] }
 thiserror = { version = "2.0" }
 time = { version = "0.3.17", features = ["serde-human-readable"] }
 tokio = { version = "1.40", features = ["rt-multi-thread", "signal", "fs"] }
-tokio-tungstenite = { version = "0.21", features = [
+tokio-tungstenite = { version = "0.24", features = [
```
hatchan, line 65: This isn't to the latest version, rather it is the version that `axum` has in `0.7`

fpx-cli/Cargo.toml
```diff
@@ -69,12 +73,5 @@ tracing-opentelemetry = { version = "0.28" }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 url = { version = "2.5" }
 
-[target.'cfg(windows)'.dependencies]
```
hatchan, line 72: Removed this, as the extra feature introduces a bunch of duplicate dependencies. I guess we needed this for some compilation issue on Windows. I think for the moment we should not support Windows until this issue is resolved in `libsql`.

PR #461, opened 2025-01-30
Playground launchpad demo workflow execute

flenter:


playground/src/main.tsx
laulauland, line 1: is the eager loading making an impact on re-renders? I'd be kind of keen to leave it as slightly more eager for now (makes things load faster in the UI) and reoptimize later

laulauland (approved):

playground/src/main.tsx
flenter: The reason why i changed this was because i  was trying to avoid getting many errors in my local setup. When the playground is loaded all data gets fetched (also for content that's not needed) & navigating on the site (like to  a workflow item) means the data gets refetched (like to the traces endpoint). 

On my local setup the traces endpoint isn't available so it's especially annoying to have all these additional requests (i don't want to get used to errors in the browser's console). I was thinking we can always change it back to "viewport" if we feel that things are slow :)

That said: I can change it back to viewport if you prefer

PR #469, opened 2025-02-05
Playground: Routes sidebar improvements

brettimus:
- Renamed all instances of `ProbedRoute` to `ApiRoute`, since we are no longer probing for routes
- Remove unnecessary fields from `ProbedRoute` and simplify Schemas, Types, and helper functions accordingly
- Add support for the http method `"TRACE"`, since that is part of the OpenAPI spec
- Remove concept of a "WS" route. Treat all routes as HTTP for now
- When present, render routes grouped by OpenAPI tags (if there are no tags, just render the list of routes)
- When present, show the route `summary` field instead of its full path

image


playground/src/components/playground/NavigationPanel/RoutesPanel/RoutesPanel.tsx
```diff
@@ -3,10 +3,52 @@ import { useNavigate } from "@tanstack/react-router";
...
+      }
+
+      return acc;
+    }, {});
+  }, [routes]);
+
+  // Get sorted tag names (TODO: respect tag order from spec instead of doing alphabetical ordering)
+  const sortedTags = useMemo(() => {
+    const tags = Object.keys(routesGroupedByTags);
+    const untagged = "untagged";
```
flenter: I guess this is fine, but was thinking that perhaps a symbol would have been a better approach

flenter (approved):
lgtm. I have one coding style nitpick, I prefer to have a newline after (sections with) return statements:

``` ts
if (!value) {
   return null;
}

return parseValue(value)
```
vs
``` ts
if (!value) {
   return null;
}
return parseValue(value)
```
I like it if sections with return statements are treated like paragraphs, but it might be good to discuss this in a larger setting.

stephlow (approved):

playground/src/components/playground/NavigationPanel/RoutesPanel/RoutesPanel.tsx
brettimus: no that is better - guarantees no clash with the api, i changed it

PR #471, opened 2025-02-05
Rip out use of Hono router in the UI for path matching logic

brettimus:
- Switch to using OpenAPI-style paths (`/users/{id}` instead of `/users/:id`)
- Remove dependency on Hono's router(s) for determining matching routes in the sidebar and extracting path parameters
- Remove Hono from package.json
- Run `playground` tests in CI


flenter (approved):
lgtm

PR #472, opened 2025-02-05
Playground: adapt to tenancy related work

laulauland:
This pluralizes the endpoints and swaps out to the factory pattern so that the apiKey could be passed


packages/embedded/src/const.ts
brettimus, line 1: we call this file `constants` in the other repos, would rename for consistency

packages/embedded/src/routes/api/reports.ts
```diff
@@ -0,0 +1,28 @@
...
+  const app = new Hono();
+  
+  // Proxy all requests to fp-services but attach a token
+  app.all("*", async (c) => {
+    const url = `${PLAYGROUND_SERVICES_URL}${c.req.path}`;
+  
+    const contentType = c.req.header("content-type");
+    const headers = new Headers();
+    // Only include the bare minimum authentication and content-type headers
+    headers.set("Authorization", `Bearer ${apiKey}`);
```
brettimus, line 14: do we need to do both of these auth methods? (Authorization Bearer and x-fpx-token)

packages/embedded/tsconfig.json
```diff
@@ -9,6 +9,9 @@
     "lib": ["ESNext"],
     "esModuleInterop": true,
     "outDir": "./dist",
+    "paths": {
+      "@/*": ["./src/*"]
```
brettimus: have you made sure this path alias works when published? i still see (in the compiled js) that the paths are "@/const", for instance

laulauland: removing this to reduce possible errors

packages/embedded/src/routes/api/reports.ts
laulauland: removed!

playground/src/routes/workflows.index.tsx
```diff
@@ -1,6 +1,6 @@
 import { workflowsQueryOptions } from "@/lib/hooks/useWorkflows";
 import { createFileRoute } from "@tanstack/react-router";
-import { NewWorkflow } from "./workflow.new";
+import { NewWorkflow } from "./workflows.new";
 
 export const Route = createFileRoute("/workflow/")({
```
brettimus: do you want to update the route here as well? otherwise the url for the UI is `/workflow` still

brettimus: same goes for the other workflow files that were renamed

brettimus (approved):
approved BUT i'd suggest renaming the playground UI routes to reflect the new filenames 

("/workflows", "/workflows/new", etc)

playground/src/routes/workflows.index.tsx
laulauland: oh forgot to run tanstack router after changing this - more changes incoming

PR #473, opened 2025-02-06
Respect the `content` prop, gate `/fp/api`, modularize the playground api client, and more!

brettimus:
- Fix playground's usage of `content`, the JSON-serialize api spec from `options.openapi.content`. So now the embedded package supports passing a spec directly to the middleware

- Adds comments in playground's `index.html` to help with testing the `content` property on the openapi options serialized into the DOM

- Add docstrings to parameters in the embedded middleware so their documentation appears in intellisense popvers
  image

- Make the `apiKey` prop optional, and if it is not present, do not mount the `/fp/api` routes inside the embedded middleware

- Change Theme default to be "system" instead of "light"

- Use a common error message parser for all `lib/api/api.ts` methods

- Add error screen to Workflows when the embedded middleware doesn't have an API key present
  image

- Add an error screen to the app when the OpenAPI spec is not fetched 
  image

### DIFF
--- a/packages/embedded/src/router.ts
+++ b/packages/embedded/src/router.ts
@@ -16,7 +16,13 @@ export function createRouter<E extends Env>(
   // We therefore remove the apiKey
   const { apiKey, fpxEndpoint, ...sanitizedOptions } = options;
 
-  app.route("/api", createApiRoutes(apiKey, fpxEndpoint));
+  if (apiKey) {
+    app.route("/api", createApiRoutes(apiKey, fpxEndpoint));
+  } else {
+    app.use("/api/*", async (c) => {
+      return c.json({ error: "Fiberplane API key is not set" }, 402);
+    });
+  }
 
   const embeddedPlayground = createEmbeddedPlayground(sanitizedOptions);
   app.route("/", embeddedPlayground);
--- a/playground/src/components/ErrorScreen.tsx
+++ b/playground/src/components/ErrorScreen.tsx
@@ -2,12 +2,12 @@ import { Icon } from "@iconify/react";
 
 export function ErrorScreen(props: {
   error: Error;
-  title: string;
-  message: string;
+  title?: string;
+  message?: string;
 }) {
   const { error: _error, message } = props;
   const title = props.title ?? "Error";
-
+  const errorMessage = message ?? "An error occurred";
   return (
     <div className="min-h-screen bg-background">
       <div className="flex flex-col items-center justify-center h-screen gap-2">
@@ -19,7 +19,7 @@ export function ErrorScreen(props: {
         />
         <div className="flex flex-col items-center gap-1">
           <h2 className="text-lg font-medium">{title}</h2>
-          <p className="text-sm">{message}</p>
+          <p className="text-sm">{errorMessage}</p>
         </div>
       </div>
     </div>
--- a/playground/src/components/ErrorScreen.tsx
+++ b/playground/src/components/ErrorScreen.tsx
@@ -1,0 +1,27 @@
+import { Icon } from "@iconify/react";
+
+export function ErrorScreen(props: {
+  error: Error;
+  title: string;
+  message: string;
+}) {
+  const { error: _error, message } = props;
+  const title = props.title ?? "Error";
+
+  return (
+    <div className="min-h-screen bg-background">
+      <div className="flex flex-col items-center justify-center h-screen gap-2">
+        <Icon
+          icon="lucide:alert-triangle"
+          width={48}
+          height={48}
+          className="text-danger"
+        />
+        <div className="flex flex-col items-center gap-1">
+          <h2 className="text-lg font-medium">{title}</h2>
+          <p className="text-sm">{message}</p>
+        </div>
+      </div>
+    </div>
+  );
+}
--- a/playground/src/components/FeatureDisabledScreen.tsx
+++ b/playground/src/components/FeatureDisabledScreen.tsx
@@ -2,12 +2,10 @@ import { Icon } from "@iconify/react";
 
 export function FeatureDisabledScreen(props: {
   error: Error;
-  title?: string;
+  title: string;
   message: string;
 }) {
-  const { error: _error, message } = props;
-  const title = props.title ?? "Feature Disabled";
-
+  const { error: _error, message, title } = props;
   return (
     <div className="min-h-screen bg-background">
       <div className="flex flex-col items-center justify-center h-screen gap-2">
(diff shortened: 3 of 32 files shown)


packages/embedded/src/router.ts
```diff
@@ -16,7 +16,13 @@ export function createRouter<E extends Env>(
   // We therefore remove the apiKey
   const { apiKey, fpxEndpoint, ...sanitizedOptions } = options;
 
-  app.route("/api", createApiRoutes(apiKey, fpxEndpoint));
+  if (apiKey) {
+    app.route("/api", createApiRoutes(apiKey, fpxEndpoint));
+  } else {
+    app.use("/api/*", async (c) => {
+      return c.json({ error: "Fiberplane API key is not set" }, 402);
```
laulauland, line 23: nit but shouldn't this be 401 instead?

brettimus: explained in slack the reasoning for 402, but i can easily change it

playground/src/components/ErrorScreen.tsx
```diff
@@ -0,0 +1,27 @@
+import { Icon } from "@iconify/react";
+
+export function ErrorScreen(props: {
+  error: Error;
+  title: string;
+  message: string;
```
brettimus: todo - make message optional and take it from the error

playground/src/components/FeatureDisabledScreen.tsx
```diff
@@ -0,0 +1,27 @@
+import { Icon } from "@iconify/react";
+
+export function FeatureDisabledScreen(props: {
+  error: Error;
+  title?: string;
+  message: string;
```
brettimus, line 6: todo - make message optional and take it from the error

playground/src/lib/api.ts
```diff
@@ -23,13 +23,43 @@ function getBasePrefix(): string {
   return mountedPath;
 }
 
+export class FpApiError extends Error {
```
brettimus: todo - make api its own module and split up this file

playground/src/lib/hooks/useOpenApiSpec.ts
```diff
@@ -12,19 +13,15 @@ export function openApiSpecQueryOptions(openapi: OpenApiContext | undefined) {
     queryKey: OPENAPI_CACHE_KEY,
     queryFn: async () => {
       if (openapi?.content) {
-        return JSON.stringify(openapi.content);
+        return openapi.content;
```
brettimus, line 16: this is the fix that made the `content` prop work correctly when passed down from the middleware

playground/src/lib/hooks/useOpenApiSpec.ts
```diff
@@ -12,19 +13,15 @@ export function openApiSpecQueryOptions(openapi: OpenApiContext | undefined) {
...
         return undefined;
       }
 
-      const response = await fetch(openapi.url);
-      if (!response.ok) {
-        throw new Error(`HTTP error! status: ${response.status}`);
-      }
-
-      return response.text();
+      const stringifiedSpec = await api.getOpenApiSpec(openapi.url);
```
brettimus, line 23: uses an api helper that will throw custom errors, that in turn the frontend knows how to render with helpful error messages

playground/src/main.tsx
```diff
@@ -36,7 +36,7 @@ declare module "@tanstack/react-router" {
 
 createRoot(rootElement).render(
   <StrictMode>
-    <ThemeProvider defaultTheme="light" storageKey="vite-ui-theme">
+    <ThemeProvider defaultTheme="system" storageKey="vite-ui-theme">
```
brettimus, line 38: default to system theme instead of light theme everywhere

playground/src/routes/__root.tsx
```diff
@@ -78,7 +92,7 @@ function ErrorBoundary({
         />
         <p className="text-lg">
           {error.message}
-          {info?.componentStack}
+          {process.env.NODE_ENV !== "production" ? info?.componentStack : null}
```
brettimus, line 94: only render the component stack in development

playground/src/routes/workflows.tsx
```diff
@@ -18,13 +21,33 @@ import { useHotkeys } from "react-hotkeys-hook";
...
     const workflowsResponse = await queryClient.ensureQueryData(
       workflowsQueryOptions(),
     );
+
     return {
       workflows: workflowsResponse.data,
     };
   },
+  errorComponent: ({ error }) => {
+    if (isFpApiError(error) && error.statusCode === 402) {
```
brettimus: todo - use an "isFeatureDisabledError" helper instead of two conditionals

playground/src/lib/api/api.ts
```diff
@@ -0,0 +1,180 @@
...
+  updateWorkflow: async (
+    id: string,
+    data: {
+      prompt: string;
+      openApiSchemaId: string;
+      summary?: string;
+      description?: string;
+    },
+  ): Promise<ApiResponse<Workflow>> => {
+    return fpFetch<ApiResponse<Workflow>>(`/api/workflows/${id}`, {
```
brettimus, line 44: `fpFetch` does a bunch of logic that was repeated across all the functions in the `api` object

- check `if (response.ok)`
- for error responses, throw custom `FpApiError` errors with additional context that the frontend can use on error screens
- default parse response as json

mies (approved):

playground/src/lib/api/api.ts
```diff
@@ -0,0 +1,169 @@
...
+    });
+  },
+
+  // TODO - Return actual response data, with a report id (once we have it)
+  createReport: async (data: {
+    traceId: string;
+    description: string;
+  }) => {
+    // NOTE - For testing the flow in the UI
+    // TODO - Remove this once ui is more finalized
```
flenter, line 130: Is this now?

playground/src/lib/api/api.ts
```diff
@@ -0,0 +1,169 @@
...
+      },
+      async () => {},
+    );
+  },
+
+  getTraceSummary: async (data: {
+    traceId: string;
+    spans: TraceDetailSpansResponse;
+  }): Promise<ApiResponse<{ summary: string }>> => {
+    // HACK - To avoid excessive requests to the API
```
flenter, line 152: Is this still needed?

laulauland (approved):

flenter (approved):

playground/src/lib/api/api.ts
brettimus: porque no

playground/src/lib/api/api.ts
brettimus: yeah let's keep it until we tackle the assistant again. it's less time to uncomment this line than to find an example response, put it in the ui, bla bla

PR #475, opened 2025-02-07
Add scratchpad functionality

flenter:
And also:
* make activeRoute the sole indicator for what route is currently selected 
* make path/method in top section of ui not editable (it's now a dropdown where you can select a path) (FP-4312)

https://github.com/user-attachments/assets/1608c905-906f-4396-ae7c-ba5c2e1d96ba


playground/src/components/mode-toggle.tsx
brettimus, line 1: heads up - i committed directly to `playground` to make this component look better (it was only serving theme selection on the settings page) 

you'll get conflicts, but please just accept the version of the file from `playground`

playground/src/components/playground/PlaygroundInput.tsx
```diff
@@ -22,36 +32,62 @@ export function PlaygroundInput({
...
-          type="text"
-          value={path}
-          onChange={(e) => {
-            handlePathInputChange(e.target.value);
-          }}
-          className="flex-grow text-xs w-full bg-transparent font-mono border-none shadow-none focus:ring-0 ml-0"
-        />
-      </div>
+      <FpDropdownMenu>
+        <FpDropdownMenuTrigger className="flex flex-grow items-center ml-2 gap-2 h-min hover:bg-muted data-[state=open]:bg-muted rounded-sm">
```
brettimus: image

this component got a little more padded than its predecessor, not sure if that was intentional

playground/src/components/playground/PlaygroundPage.tsx
```diff
@@ -26,6 +27,9 @@ export const PlaygroundPage = () => {
     "sidePanel",
     "shouldShowTopNav",
   );
+  const hasRoutes = useStudioStoreRaw(
+    useShallow((state) => state.appRoutes.length > 0 && state.activeRoute),
```
brettimus, line 31: note to self: `hasRoutes` also captures the loading state, no page content will show while routes are being fetched

playground/src/components/playground/PlaygroundInput.tsx
brettimus: for reference: 
image

playground/src/components/playground/PlaygroundInput.tsx
```diff
@@ -22,36 +32,66 @@ export function PlaygroundInput({
...
+                  )}
+                >
+                  <div
+                    className={cn(
+                      "flex-grow grid grid-cols-[2.5rem_auto] items-center gap-1 px-1 rounded-md",
+                    )}
+                  >
+                    <Method method={route.method} className="ml-1 text-xs" />
+                    <div className="flex w-full my-2">
+                      <div className="flex-initial text-xs w-full bg-transparent font-mono border-none shadow-none focus:ring-0 ml-0 disabled:cursor-text disabled:bg-muted py-0">
```
brettimus: this will be tricky but we probably want to truncate this somehow, especially as the number of query params gets large

image

playground/src/components/playground/PlaygroundInput.tsx
brettimus: okay i tightened up padding and vertically centered the text in a commit, let me know what you think

image

playground/src/components/playground/PlaygroundInput.tsx
brettimus: okay this is what i went with: 
image

brettimus (commented):
first round of review, some notes to self and then screenshots of things i modified. 

still making my way through, just wanted to put these comments up in case you had thoughts @flenter

playground/src/components/playground/PlaygroundInput.tsx
brettimus: maybe we do a ...... tooltip?!

playground/src/utils/http-method-string.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Returns a string representation of the HTTP method.
+ *
+ * In practice, shortens the method name to be 3-5 characters.
+ */
+export function getHttpMethodString(method: string) {
```
brettimus, line 6: added this to never have more than 5 chars in the Method component, or when we show an HTTP method elsewhere

playground/src/components/playground/PlaygroundInput.tsx
```diff
@@ -22,35 +32,79 @@ export function PlaygroundInput({
...
+                  >
+                    <div
+                      className={cn(
+                        "flex-grow grid grid-cols-[2.5rem_auto] items-center gap-1 px-1 rounded-md",
+                      )}
+                    >
+                      <Method method={route.method} className="ml-1 text-xs" />
+                      <div className="flex w-full my-2">
+                        <div className="flex-initial text-xs w-full bg-transparent font-mono border-none shadow-none focus:ring-0 ml-0 disabled:cursor-text disabled:bg-muted py-0">
+                          {route.summary || route.description || route.path}
```
brettimus, line 98: i removed `ApiRoute["title"]` as OpenAPI only specifies "summary" and "description" fields 

for our purposes "summary" is the route title

wondering if we also wanna show description in this dropdown if it's available?

image

playground/src/components/playground/PlaygroundInput.tsx
flenter: The reason i made it take up a bit more space is that when you change from get/post etc i don't like it if the url jumps

playground/src/components/playground/PlaygroundPage.tsx
flenter: We should probably add an empty state/loading screen or something like that, but i didn't want to add that to this pr

playground/src/components/playground/PlaygroundInput.tsx
flenter: Perhaps a designer can make a nice suggestion here. This dropdown is a bit clunky

playground/src/components/playground/PlaygroundInput.tsx
brettimus: gotcha. i added a min width to avoid jumps

playground/src/components/playground/store/utils.ts
```diff
@@ -97,7 +97,7 @@ export function resolvePathWithParameters(
   pathParams: KeyValueParameter[],
 ) {
   return pathParams.reduce((acc, param) => {
-    return acc.replace(`{${param.key}}`, param.value || param.key);
+    return acc.replace(`{${param.key}}`, encodeURIComponent(param.value));
```
brettimus, line 100: nice

brettimus (approved):
woooo good improvements in here

PR #478, opened 2025-02-10
Fp 4331 add debug logging to the embedded middleware

mellowagain:


packages/embedded/src/routes/api/assistant.ts
```diff
@@ -1,26 +1,37 @@
...
     }
 
-    return fetch(url, {
+    const result = fetch(url, {
       method: c.req.method,
       headers,
       body: c.req.raw.body,
     });
+
+    logIfDebug(c.get("debug"), "proxied request sent off. response:", result);
```
brettimus: so for this and also in `workflows`, the result here is a Promise, so logging the promise isn't enormously helpful. i add some logging upstream (in `middleware.ts`) to signify when the request completed

brettimus (approved):
awesome. i moved some things, made the logIfDebug helper polymorphic, and then added vitest to do some very rudimentary tests

feel free to look at the commits i added! but i'm going to merge for now

PR #479, opened 2025-02-10
Prepare embedded package for release

brettimus:
- Rename `@fiberplane/embedded` to `@fiberplane/hono` and publish as `0.4.0`
- Rename `createMiddleware` to `createFiberplane`
- Still export `createMiddleware`, but with an `@deprecated` flag ((since we already told a handful of people to use `createMiddleware`))
- Fix type issues with exported `createFiberplane` (simplified type definition of `createFiberplane`)
- When `apiKey` is not passed to the options, fall back to the `FIBERPLANE_API_KEY` found on `c.env`
- Add a few more debug logs

## A note on testing 

To test the resolution of the type-related issue, I had to be sure to `pack` the embedded package and install it in another api locally. Linking directly to the monorepo caused type clashes because the monorepo's installed version of Hono and the test api's version of Hono.


laulauland (approved):

packages/embedded/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fiberplane/embedded",
```
laulauland: also maybe rename this to just `fiberplane`?

brettimus: per conversation, changing to `@fiberplane/hono`

PR #486, opened 2025-02-12
Use openapi data structure

flenter:
This PR removes our own custom implementation of parsing/typing the OpenAPI spec that we used in parts of the codebase. The openapi spec is now a first class citizen .

I've also:
* added @tanstack/react-virtual to the `RoutesPanel` so when testing with big openapi specs (like the cloudflare openapi spec), some interactions would sometimes be very slow
* removed `immer` from the `zustand` store (this was done because the dereferenced api spec would cause `maximum call stack size exceeded` errors (possibly due to circular references in the cloudflare openapi spec)


flenter:
Closing this in favor of #497

PR #489, opened 2025-02-13
Support (local) Tracing in Playground

brettimus:
## TODOs

- [x] Redact neon headers that contain sensitive information (sql queries, etc)
- [x] ~~`waitUntil` the fetches inside the custom span exporter~~ _not necessary_

### Overview

Adds necessary changes to 

- Prevent transmission of sensitive information outside of local mode
- Make certain functionality work in Deno (valtown)

### Library Behavior Changes

**In production:**

- Do not send environment variables
- Do not record request bodies
- Do not record SQL queries
- Do not record result of CF binding calls
- Redact sensitive header values (configurable)
- Redact sensitive query parameters (configurable)

**General**

- Break studio compatibility! I.e., do **_not_** send an updated list of routes to the otel endpoint

### "local mode" vs production

The client library needs to be aware of whether it is in production. In production, we will NOT send certain sensitive data (see below).

To determine whether we're in "local" mode, the library first checks:

- Whether the `FIBERPLANE_ENVIRONMENT` env var `=== "local"`

If there is no `FIBERPLANE_ENVIRONMENT` set, then we check if `FIBERPLANE_OTEL_ENDPOINT` contains `localhost`.

### Env var changes

The names of several environment variables that are used for configuration has changed. The old ones will still work, but new ones will take precedence over the olds ones.

#### Name changes

- `FPX_ENDPOINT` (_deprecated_) => `FIBERPLANE_OTEL_ENDPOINT`
- `FPX_LOG_LEVEL` (_deprecated_) => `FIBERPLANE_OTEL_LOG_LEVEL`
- `FPX_SERVICE_NAME` (_deprecated_) => `FIBERPLANE_SERVICE_NAME`
- `FPX_AUTH_TOKEN` (_deprecated_) => `FIBERPLANE_OTEL_TOKEN`

### Custom SpanExporter

The OpenTelemetry libraries broke when we added auth headers to the out-of-the box exporter, so I needed to write a custom span exporter. This is in the client library package under `exporter.ts`

## `@fiberplane/hono` embedded middleware

- Add support for the new env var names (listed above)
- Add support for reading env vars from Deno environments
- Turn off AI summaries

## Playground UI

- Send `x-fpx-ignore` header with internal API calls, to avoid spamming the tracing page with calls to our external API.
- Manually filter out preflight requests from the result set of traces
- Turn off AI summaries


packages/client-library-otel/package.json
```diff
@@ -4,11 +4,11 @@
   "author": "Fiberplane<info@fiberplane.com>",
   "type": "module",
   "main": "dist/index.js",
-  "version": "0.7.0",
+  "version": "0.8.0-canary.2",
   "dependencies": {
     "@opentelemetry/api": "~1.9.0",
-    "@opentelemetry/exporter-trace-otlp-http": "^0.52.1",
-    "@opentelemetry/exporter-trace-otlp-proto": "^0.52.1",
+    "@opentelemetry/exporter-trace-otlp-http": "0.35.0",
```
brettimus, line 11: downgraded these because they were the versions released with `@opentelemetry/api@1.9.0` 

pinning them because they are experimental and change their apis a lot

packages/client-library-otel/src/exporter.ts
```diff
@@ -0,0 +1,124 @@
...
+      return value;
+    });
+
+    const params: RequestInit = {
+      method: "POST",
+      headers: this.headers,
+      body,
+    };
+
+    if (isWrapped(webStandardFetch)) {
```
brettimus: alternative to throwing an error would be to unwrap and rewrap then await the Promise<Response> 

but i don't think we'll hit that need, since we freeze `fetch` at the top of the file, before we'd do any runtime monkeypatching

packages/client-library-otel/src/config/config.ts
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * We want to minimize configuration, so instead, we've chosen to ignore the following headers by default.
+ *
+ * In practice, the library only redacts their values when running in "production" mode.
+ */
+export const DEFAULT_REDACTED_HEADERS = [
+  "authorization",
+  "cookie",
+  "set-cookie",
+  "x-api-key",
```
flenter, line 39: maybe also add the following to the list?

* `x-auth-token`
* `x-csrf-token`
* `x-session-id`

packages/client-library-otel/src/config/index.ts
```diff
@@ -0,0 +1,15 @@
+export { resolveConfig } from "./config";
+export type { FpxConfigOptions } from "./types";
```
flenter: I guess it's OK to leave this in for now, but it's a reference to the old `fpx` branding (also applies elsewhere)

packages/client-library-otel/src/config/resolvers.ts
```diff
@@ -0,0 +1,110 @@
...
+      lowercaseEndpoint.includes("::1")
+    );
+  }
+}
+
+/**
+ * Get the OTLP endpoint from the environment variables.
+ */
+export function getOtelEndpoint(env: FpHonoEnv) {
+  return getFromEnv(env, [
```
flenter, line 70: Would be nice if there's a deprecation warning when using the `FPX` env var name (same applies elsewhere)

packages/client-library-otel/src/exporter.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  send(
+    // biome-ignore lint/suspicious/noExplicitAny: items really can be anything
+    items: any[],
+    onSuccess: () => void,
+    onError: (error: OTLPExporterError) => void,
+  ): void {
+    const exportMessage = createExportTraceServiceRequest(
+      items,
+      /** useHex: */ true,
```
flenter, line 84: This probably should be:
`{ useHex: true }`

packages/client-library-otel/src/exporter.ts
```diff
@@ -0,0 +1,135 @@
...
+    this._export(items)
+      .then(() => {
+        resultCallback({ code: ExportResultCode.SUCCESS });
+      })
+      .catch((error: ExportServiceError) => {
+        resultCallback({ code: ExportResultCode.FAILED, error });
+      });
+  }
+
+  // biome-ignore lint/suspicious/noExplicitAny: items really can be anything
```
flenter: Why type it to any? The code really only accepts `Array<ReadableSpans>` (this also applies elsewhere).

I tried this change in the code and at least typescript isn't complaining about it (same applies elsewhere in this file)

packages/playground/src/components/FeatureDisabledScreen.tsx
```diff
@@ -18,6 +22,14 @@ export function FeatureDisabledScreen(props: {
         <div className="flex flex-col items-center gap-1">
           <h2 className="text-lg font-semibold text-foreground">{title}</h2>
           <p className="text-foreground/70 text-center max-w-md">{message}</p>
+          <Button
```
flenter: Why not make this a link?

packages/playground/src/components/traces/TracesList/TraceElement.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <span>
+              <Link
+                to="/traces/$traceId"
+                params={{ traceId: trace.traceId }}
+                className={cn(
+                  "inline-flex items-center gap-0.5",
+                  "transition-colors hover:underline hover:text-foreground",
+                )}
+              >
+                View Trace Details{" "}
```
flenter: Is the " " space needed? Given the gap-0.5?

packages/playground/src/routes/traces.$traceId.tsx
```diff
@@ -36,12 +37,11 @@ import {
 import { useQueryClient } from "@tanstack/react-query";
 import { createFileRoute } from "@tanstack/react-router";
 import { useEffect, useMemo, useState } from "react";
-import { useHotkeys } from "react-hotkeys-hook";
+// import { useHotkeys } from "react-hotkeys-hook";
```
flenter: 👋

flenter (commented):
Some smaller feedback. Will play with it/look again tomorrow.

packages/client-library-otel/src/config/config.ts
brettimus: f32153c31f2d5c2d009127614ebc7603a868ea5b

packages/client-library-otel/src/exporter.ts
brettimus: yeah i had to downgrade the otel libraries, and this version only accepts a single boolean (useHex)

packages/client-library-otel/src/exporter.ts
brettimus: good catch!

brettimus: bf81727c8f1b2a13048fe372d52a2a26e04afe23

packages/playground/src/components/traces/TracesList/TraceElement.tsx
brettimus: removed this link

packages/client-library-otel/src/config/index.ts
brettimus: gotcha. i removed

brettimus: er. renamed

packages/client-library-otel/src/config/resolvers.ts
brettimus: good point. but in that case i should've written this as a chain of "resolvers" (functions), where the deprecated one logs a warning

 probably a little too late to do that, but i can file a ticketke

brettimus: https://linear.app/fiberplane/issue/FP-4417/log-a-warning-when-someone-uses-a-deprecated-env-var-like-fpx-endpoint

packages/playground/src/components/FeatureDisabledScreen.tsx
brettimus: good point. original idea was to use a "go back" button, which i've now implemented as well

packages/playground/src/components/traces/TracesList/TraceElement.tsx
```diff
@@ -43,43 +50,67 @@ export function TraceElement({ trace }: { trace: Trace }) {
     return null;
   }
 
-  const incomingRequest = waterfall[0];
+  // HACK - Need a better typesafe way to get the incoming request span
+  const incomingRequest = waterfall[0] as SpanWithVendorInfo;
```
flenter, line 54: There are no checks to see if there actually are any items in the array, is there?

packages/playground/src/components/traces/TracesList/TraceElementHeader.tsx
```diff
@@ -0,0 +1,195 @@
...
+          className="flex items-center justify-around h-7
+       pr-3 pl-1"
+        >
+          <Icon icon="lucide:chevrons-up-down" className="w-4 h-4" />
+        </DivWithHover>
+        <div
+          className={cn(
+            "grid",
+            "grid-cols-[1fr]",
+            // "items-center",
```
flenter: 👋

packages/playground/src/components/traces/TracesList/TraceElementHeader.tsx
```diff
@@ -0,0 +1,195 @@
...
+    <div
+      {...props}
+      className={cn("group-hover:bg-primary/10 cursor-pointer", className)}
+    >
+      {children}
+    </div>
+  );
+};
+
+import { StatusCode } from "@/components/StatusCode";
```
flenter: Can you move the imports to the top of the file?

flenter (approved):
A couple of small things but LGTM!

PR #496, opened 2025-02-17
Change workflow data structure, tweak vite.config.ts, publish new version

laulauland:
This makes sure that the playground can work with the new data structure of workflows.

It also tweaks the `vite.config.ts` file to make the dev server setup centralized.

TODO:
- [ ] publish the new version once approved before merging


packages/playground/src/routes/workflows.$workflowId.tsx
```diff
@@ -120,7 +124,7 @@ function WorkflowDetail() {
 
   if (loadingError || parsingError) {
     return (
-      <div className="grid place-items-center h-full">
+      <div className="grid h-full place-items-center">
```
flenter, line 127: Why is the order changed? Is that because of an extension? If so, what are you using?

flenter (approved):

packages/playground/src/routes/workflows.$workflowId.tsx
laulauland: yep it's [Headwind](https://github.com/heybourn/headwind)

PR #500, opened 2025-02-18
Add blogpost async task 2b

Nlea:
update image and video


www/src/main.css
```diff
@@ -30,9 +30,7 @@
   --sl-color-text-accent: var(--fg-brand);
 
   /* Fix for misaligned elements in nav/header */
-
-  --sl-nav-pad-y: 0.55rem;
-
+  --sl-nav-pad-y: 0.625rem;
```
brettimus: was this intentional? makes me a little uneasy to see a toplevel variable tweaked 👀

Nlea: @brettimus good question I ca change it back

Nlea: I changed it

brettimus (approved):

www/src/main.css
brettimus: thanks! approved

PR #501, opened 2025-02-18
Support ~boolean parameters~ checkboxes

flenter:
It looks amazing:

image

Also: i've split up the KeyValueFormRowValue component.

### DIFF
--- a/packages/playground/src/components/playground/KeyValueForm/KeyValueFormRowValue.tsx
+++ b/packages/playground/src/components/playground/KeyValueForm/KeyValueFormRowValue.tsx
@@ -14,13 +14,14 @@ import {
 import { Input } from "@/components/ui/input";
 import { FpLabel } from "@/components/ui/label";
 import { FpRadioGroup, FpRadioGroupItem } from "@/components/ui/radio-group";
-import { isSupportedSchemaObject } from "@/lib/isOpenApi";
-import { cn } from "@/utils";
+import { isSupportedSchemaObject, type SupportedSchemaObject } from "@/lib/isOpenApi";
+import { cn, isMac } from "@/utils";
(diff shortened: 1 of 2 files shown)


packages/playground/src/components/playground/KeyValueForm/KeyValueFormRowValue.tsx
```diff
@@ -231,3 +147,248 @@ export function KeyValueFormRowValue(props: KeyValueRowValueProps) {
...
+}
+
+function KeyValueBoolean(props: {
+  value: string;
+  onChangeValue(value: string): void;
+}) {
+  const { value, onChangeValue } = props;
+
+  return (
+    <Checkbox
```
laulauland, line 388: can we still add a `true/false` label that would mirror the state? We use checkbox for a few things in the UI so would just make it unambiguous about what kind of data this refers to

packages/playground/vite.config.ts
```diff
@@ -91,7 +91,7 @@ function injectPlaygroundDevConfig(): Plugin {
         mountedPath: "/",
         openapi: {
           // Comment out the url to use the sample OpenAPI spec
-          url: "http://localhost:7676/api/openapi.json",
+          url: "http://localhost:8787/openapi.json",
```
laulauland: 👀

laulauland (approved):

laulauland:
1 small comment but LGTM

packages/playground/src/components/playground/KeyValueForm/KeyValueFormRowValue.tsx
flenter: Sorry, I somehow missed this comment, would you like to display the text "true/false" next to a checkbox?

PR #509, opened 2025-02-25
Workflows ui tweak

flenter:
https://github.com/user-attachments/assets/538933dd-f3b6-44af-b864-47036343cbcd

This pr:
* updates the create workflow form 
* adds interactivity to the steps list
* fixes rendering issue with settings (where the UI would not always update)
* fixes active state on the to NavLinks

### DIFF
--- a/packages/playground/src/components/WorkflowDetail/WorkflowDetail.tsx
+++ b/packages/playground/src/components/WorkflowDetail/WorkflowDetail.tsx
@@ -16,7 +16,6 @@ export function WorkflowDetail() {
     </div>
   );
 
-  console.log("workflow", workflow);
   return (
     <div className="overflow-y-auto h-full">
       <ListSection title={workflow.summary}>
@@ -66,7 +65,7 @@ export function WorkflowDetail() {
                 </ListSection>
               </div>
             </ListSection>
-            <ListSection title="Steps" contentClassName="px-0 pr-2">
+            <ListSection title="Steps" contentClassName="px-1 pr-2">
               <div>
                 {workflow.steps.map((step, index) => (
                   <StepperItem
--- a/packages/playground/src/components/WorkflowDetail/WorkflowDetail.tsx
+++ b/packages/playground/src/components/WorkflowDetail/WorkflowDetail.tsx
@@ -16,6 +16,7 @@ export function WorkflowDetail() {
     </div>
   );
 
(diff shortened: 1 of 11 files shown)


packages/playground/src/components/WorkflowDetail/WorkflowDetail.tsx
```diff
@@ -16,6 +16,7 @@ export function WorkflowDetail() {
     </div>
   );
 
+  console.log("workflow", workflow);
```
laulauland: ✌️

packages/playground/src/routes/workflows.new.tsx
```diff
@@ -21,36 +22,36 @@ export function NewWorkflow() {
     console.error("No OpenAPI spec found");
   }
 
-  const handleSubmit = () => {
-    createWorkflow({
+  const handleSubmit = useHandler(() => {
```
laulauland: shouldn't we recreate this callback when user story or (unlikely) open api schema changes?

laulauland (approved):
two quick comments but looks good

packages/playground/src/routes/workflows.new.tsx
flenter: `useHandler` doesn't actually do that much here, given that the WorkflowPrompt component gets re-rendered when the story changes and the `useHandler` hook keeps the function signature stable even though the function is recreated due to a re-render or a change to the `userStory`. Using `useHandler` the handleSubmit function always refers to the latest values for both the openapi spec as well as the prompt/`userStory`.

I think we may want to rely on this hook in our codebase a bit more often (I typically prefer it above `useCallback` or redefining inline functions). The hook keeps the function signature stable can help avoid re-renders (and issues with `useEffect` getting triggered). 

For reference, here's the code for the `useHandler` hook:

``` ts 
import { useCallback, useRef } from "react";

const noDeps: Array<void> = [];

// biome-ignore lint/suspicious/noExplicitAny: not aware of a better solution in TS
export function useHandler<Handler extends (...args: Array<any>) => any>(
  handler: Handler,
): Handler {
  const handlerRef = useRef(handler);
  handlerRef.current = handler;

  // @ts-ignore
  return useCallback((...args) => handlerRef.current(...args), noDeps);
}
```

That said: I think i'm going to remove it as it doesn't really help preventing useless re-renders here.

PR #511, opened 2025-02-27
Support various playground local development scenarios

brettimus:
This is a PR that adds:

1. Some configuration to the playground SPA that makes local development easier. 
2. Improvement to `@fiberplane/hono`: Add `tsc-alias` so we no longer need to specify `.js` extensions in imports
3. Pass in configuration to `@fiberplane/hono` to control which instance of fp-services it makes requests against

It also removes an example from the `examples` folder that is a little crufty.

> PR description will get updated as we build consensus on a few different ways of doing things

## Proposed Default for `pnpm dev` in the playground package

When you run `pnpm dev`, the playground spa will assume:

- You have a local API running on `http://localhost:8787`
- You have the fiberplane middleware mounted on `/fp`
- Your openapi spec is available at `/openapi.json` on the local API

## Proposed Local Development Flow for Sample APIs

Say you want to run the SPA against a locally running instance of a sample API like `fashion`, or the mock music api. Just run:

```sh
cd packages/playground
pnpm dev:example
```

## Proposed Local Development Flow for Dog-fooding

Say you want to run the SPA against a locally running instance of `fp-services`:

```sh
cd packages/playground
pnpm dev:dogfood
```


packages/playground/index.html
```diff
@@ -4,183 +4,10 @@
     <meta charset="UTF-8" />
     <link rel="icon" type="image/svg+xml" href="/fp-logo-fav.svg" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Fiberplane Playground</title>
+    <title>(Local) Fiberplane Playground</title>
   </head>
   <body>
     <div id="root"></div>
     <script type="module" src="/src/main.tsx"></script>
-    <script>
```
brettimus, line 12: we want to rely on vite to inject this code, so should remove it from `index.html`

flenter (approved):

mies (approved):

packages/playground/vite.config.ts
```diff
@@ -43,19 +50,23 @@ export default defineConfig({
   server: {
     port: 6660,
     proxy: {
-      "/w": {
+      "/w/*": {
```
flenter: Why is the asterisk required? I think it should be the same/similar to what we do for `/api`

flenter (changes requested):
For me the proxying of `/w` calls didn't work correctly when running the services api locally

packages/playground/vite.config.ts
flenter: I think `/w/` is probably what works best here.

mies (approved):

flenter (approved):

PR #513, opened 2025-03-03
feat: added hono-openapi example

MathurAditya724:
- [x] Copied `zod-openapi` example and converted it for `hono-openapi`, Only 2 files are changed -
  - package.json
  - index.ts

Let me know if any more changes are required.

Quick question: Would we need to copy this example in the `create-honc-app` to add this in the CLI?

### DIFF
--- a/examples/hono-openapi/.dev.vars.example
+++ b/examples/hono-openapi/.dev.vars.example
@@ -1,0 +1,1 @@
+FPX_ENDPOINT=http://localhost:8788/v1/traces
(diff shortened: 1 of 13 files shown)


brettimus:
right on! thank you! as of today we just retired the local studio and are working with a new piece of middleware that embeds a studio-like interface in the api itself. going to merge this for now then make some tweaks. will tag you in the PR!

brettimus (approved):

PR #517, opened 2025-03-04
Add partition keys to workflows

laulauland:
This adds UI and package changes to allow the user to set partition keys in order to access workflows. Partition keys is a stopgap solution to identify the API user using the playground. 

Note: don't merge this until the Fiberplane API changes are merged and deployed

### DIFF
--- a/packages/playground/src/components/playground/store/slices/settingsSlice.ts
+++ b/packages/playground/src/components/playground/store/slices/settingsSlice.ts
@@ -286,10 +286,7 @@ export const settingsSlice: StateCreator<
         const state = { ...initialState };
         state.partitionKey = key;
 
-        localStorage.setItem(
-          SETTINGS_STORAGE_KEY,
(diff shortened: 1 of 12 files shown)


packages/playground/src/components/playground/store/slices/settingsSlice.ts
```diff
@@ -270,6 +278,22 @@ export const settingsSlice: StateCreator<
...
+    setPartitionKey: (key: string) =>
+      set((initialState: StudioState): StudioState => {
+        const state = { ...initialState };
+        state.partitionKey = key;
+
+        localStorage.setItem(
+          SETTINGS_STORAGE_KEY,
+          JSON.stringify({
+            ...state,
+            partitionKey: state.partitionKey,
```
flenter: Why not simply stringify the state?

flenter (approved):
one minor thing but apart from that: lgtm

PR #522, opened 2025-03-05
Playground auth

brettimus:


packages/fiberplane-hono/src/routes/api/auth.ts
```diff
@@ -0,0 +1,100 @@
...
+        headers: {
+          Authorization: `Bearer ${apiKey}`,
+          "Content-Type": "application/json",
+          Cookie: `session=${sessionKey}`,
+        },
+      });
+    }
+
+    return c.json({ success: true });
+  });
```
brettimus: not yet implemented but this was my first pass at the logout api route inside the embedded

PR #537, opened 2025-03-24
make command bar work in traces ui by moving it to layout

mellowagain:
fixes FP-4421


packages/playground/src/Layout/Layout.tsx
```diff
@@ -36,12 +38,125 @@ NavButtonComponent.displayName = "NavButtonComponent";
...
+    "shouldShowTopNav",
+    "setShortcutsOpen",
+    "setActiveRequestsPanelTab",
+    "visibleRequestsPanelTabs",
+    "setCommandBarOpen",
+  );
+
+  const formRef = useRef<HTMLFormElement>(null);
+
+  useHotkeys(
```
brettimus: okay so a lot of these hotkeys are specific to the playground page - like 

- requestSubmit
- go to docs
- go to request params
- etc

might be good to move those hooks back to the playground index page

packages/playground/src/components/playground/PlaygroundPageContent/PlaygroundPageContent.tsx
```diff
@@ -38,108 +27,11 @@ export const PlaygroundPageContent: React.FC = (_props) => {
...
-
-  useKeySequence(
-    ["g", "d"],
-    () => {
-      if (visibleRequestsPanelTabs.includes("docs")) {
-        setActiveRequestsPanelTab("docs");
-      }
-    },
-    { description: "View Route Docs", ignoreSelector: "[contenteditable]" },
+  const { commandBarOpen, setCommandBarOpen } = useStudioStore(
```
brettimus: okay another thing: the `<CommandBar` component is still only rendered on this page, so we may need to move it up into the layout component itself, so that each page still has the component that can render the command bar itself

brettimus (approved):

PR #541, opened 2025-03-26
Stabilizing the agents playground

laulauland:
This ensures that the agents playground can be installed and used with just `@fiberplane/agents` package.

> [NOTE]
> This is still a heavy work in progress that makes certain assumptions about the structure of the agents project.

Major changes:

1. Uses `fiberplane` helper to patch the worker entrypoint `fetch` to get the agents and their instances (as opposed to the vite plugin).
2. Has a bunch of UI refactors (added zustand as the general state/store, moved the registration/listening to SSE events to the agents detail component so the events are now more frequently listened to, streamlined the events tab, made the event details page more responsive, added more obvious mouse over effect to the (Agent) card, added empty states in several places and more)


packages/agents-playground/src/hooks/useAgentDB.ts
```diff
@@ -0,0 +1,12 @@
+import type { DatabaseResult } from "@/types";
+import { useQuery } from "@tanstack/react-query";
+
+export function useAgentDB(namespace: string, instance: string) {
+  return useQuery({
+    queryKey: ["agent_db", namespace, instance],
```
laulauland: nit: can we set the key as constant

packages/agents-playground/src/hooks/useSSE.ts
```diff
@@ -0,0 +1,364 @@
...
+    close,
+    clearEvents,
+  };
+}
+
+/**
+ * Legacy hook that provides the same API as the original useSSE hook
+ * @deprecated Use useSSEWithEvents instead
+ */
+export function useSSE<T = unknown>(
```
laulauland: can we remove the old `useSSE` if it's not being used altogether

packages/agents-playground/src/hooks/useListAgents.ts
```diff
@@ -0,0 +1,9 @@
+import type { ListAgentsResponse } from "@/types";
+import { useQuery } from "@tanstack/react-query";
+
+export function useListAgents() {
+  return useQuery<ListAgentsResponse>({
+    queryKey: ["list_agents"],
```
laulauland: nit: same for the query key as earlier

packages/agents-playground/src/components/ui/spinner.tsx
laulauland, line 1: empty file?

laulauland (commented):
couple of small comments but looks good

flenter (approved):

PR #542, opened 2025-03-26
Register decorated Agent classes as namespaces

flenter:
And log a warning if we encounter durable objects that have not been decorated (or more precise when we can't correlate a decorated class with a durable object binding)

Also improved the empty state of the admin console slightly:
localhost_6660_ (1)

### DIFF
--- a/examples/simple-agent/src/server.ts
+++ b/examples/simple-agent/src/server.ts
@@ -9,14 +9,14 @@ import {
   generateId,
   streamText,
 } from "ai";
-import { Fiber, fiberplane } from "../../../packages/agents/dist";
+import { Fiber, fiberplane } from "@fiberplane/agents";
 import { executions, tools } from "./tools";
 import { processToolCalls } from "./utils";
 
 // Environment variables type definition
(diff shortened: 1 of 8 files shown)


examples/simple-agent/src/server.ts
```diff
@@ -10,6 +9,7 @@ import {
   generateId,
   streamText,
 } from "ai";
+import { Fiber, fiberplane } from "../../../packages/agents/dist";
```
laulauland: should use `@fiberplane/agents` here

packages/agents/src/agentInstances.ts
laulauland, line 1: nit: can we use kebab case for file names in `agents`:  `agent-instances.ts`

laulauland (approved):
2 small nits, otherwise LGTM

packages/agents/src/agentInstances.ts
flenter: i would prefer to follow the conventions that we see elsewhere in the repo.

PR #556, opened 2025-04-03
Agents playground: add send event

flenter:
This PR adds ws_send and the broadcast event. 

It also has an initial implementation of the design, updates tab design, adds resize panel, streamlines events list and adds "details" rendering for several of the events 

I want to change the way we parse events in the application, but this can come later. The store should do more around extracting details from events (as many events are strings containing a json payload which could contain another encoded string/payload). But this PR has been brewing long enough!


packages/agents-ui/src/store/slices/ui.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export type UIActions = {
+  toggleCombineEvents: () => void;
+  toggleEventCategory: (category: EventCategory) => void;
+  resetEventCategories: () => void;
+  unselectAllEventCategories: () => void;
+};
+
+// UI slice
+export const uiSlice = combine<UIState, UIActions>(
```
laulauland, line 71: arguably we could serialize this into the URL query params (there can be a lot of state in this that a user might want to "refresh" in connection hiccups). Not for now but maybe a mental note for later

packages/agents-ui/src/store/slices/agents.ts
```diff
@@ -0,0 +1,253 @@
...
+        let combinedEvents = [
+          ...(instanceDetails.combinedEvents || EMPTY_COMBINED_EVENTS),
+        ];
+        const knownBroadcastEvents = {
+          ...(instanceDetails.knownBroadcastEvents || {}),
+        };
+
+        // Handle broadcast event combining
+        if (event.type === "broadcast") {
+          try {
```
laulauland: to somewhat flatten this out maybe makes sense to move it to the `tryCatch` function?

examples/simple-agent/src/server.ts
```diff
@@ -33,13 +33,11 @@ interface MemoryState {
...
 export const agentContext = new AsyncLocalStorage<ChatClient>();
 
-export { ChatClient };
-
 /**
  * Chat Agent implementation that handles real-time AI chat interactions
  */
 @Fiber()
-class ChatClient extends AIChatAgent<Env, MemoryState> {
+export class ChatClient extends AIChatAgent<Env, MemoryState> {
```
laulauland, line 40: are you sure this is correct exporting? I had the runtime complain to me about the decorator preceding `export` keyword

laulauland (approved):
couple of comments but looks good!

packages/agents-ui/src/store/slices/ui.ts
flenter: In our local development things are a bit funky as the vite plugin is proxying the events stream. But for normal use, when the event stream gets disconnected (perhaps because some things are restarting): there's a reconnect button now

examples/simple-agent/src/server.ts
flenter: oh interesting. I didn't see that warning/error. I changed this at some point to try and fix issues with me needing to manually restart the server when I updated the @fiberplane/agents package. This change didn't resolve this issue but I kept it in because i didn't like that we were exporting the class before the actual class definition.

I will add back the `export { ChatClient };` line but below the class definition

packages/agents-ui/src/store/slices/agents.ts
flenter: I'm not sure. I think it is indeed code that would be well suited to extract out/refactor (I would like to move things to a separate function(s)). The `tryCatch` isn't really needed here because i can swap the `parse` call for a `safeParse` which basically does a similar thing here without adding another function call.

PR #570, opened 2025-04-08
Add the ability to add custom route for agents playground

laulauland:
Currently the agents playground is always available only on the `/fp` route. This adds the ability for the users to add their own route prefix.

It also refactors the `agents` source code a little bit so that everything is not mushed into a single `index.tsx` file


flenter (approved):
🎉

oscarvz (approved):

PR #572, opened 2025-04-09
@fiberplane/agents: Type events more strictly

flenter:
This adds a bit more type information to the agents library and moves the validation/parsing of the payload to the schema where it's parsed once (as opposed to parsing it in the components)

### DIFF
--- a/packages/agents-ui/src/types.ts
+++ b/packages/agents-ui/src/types.ts
@@ -1,4 +1,5 @@
 import type { QueryClient } from "@tanstack/react-query";
+import type { ToolCall, ToolResult } from "ai";
 import { z } from "zod";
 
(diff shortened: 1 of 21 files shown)


flenter:
We're now rendering websocket/broadcast messages a bit better.
image

packages/agents-ui/src/types.ts
```diff
@@ -345,3 +354,270 @@ export interface HttpResponsePayload {
   body?: unknown;
   [key: string]: unknown;
 }
+
+// // Event types
+// export const agentEventTypeSchema = z.enum([
```
laulauland: delete? 👀

laulauland (approved):

PR #588, opened 2025-04-23
Add `withInstrumentation` mixin and remove the `Observed` decorator

flenter:
This pr adds an observed mixin, example use:

```
import { withInstrumentation } from "@fiberplane/agents";

type MemoryState = Record<string, string>;

export class Example extends AIChatAgent<Env, MemoryState> {

   initialState = {};

  onStart() {
    console.log("Example agent started");
  }

  async onRequest(request: Request) {
    return new Response(
      JSON.stringify({
        url: request.url,
      }),
      {
        headers: { "Content-Type": "application/json" },
        status: 200,
      },
    );
  }
}

export const MyExample = withInstrumentation(Example)
```
The pr also removes the `@Observed` decorator making it a major breaking change to the API.


packages/agents/src/routing.tsx
```diff
@@ -0,0 +1,180 @@
...
+    request: Request,
+    env: E,
+    ctx: ExecutionContext,
+  ) => Promise<Response>,
+  options?: FiberplaneEntryWrapperOptions,
+) {
+  const fpApp = createFpApp(options?.customPath);
+
+  return async function fetch(request: Request, env: E, ctx: ExecutionContext) {
+    // console.log("[FP] Request URL:", request.url);
```
laulauland: 👀

laulauland (approved):
Looks good!

Not sure if I'm a big fan of `ObservedMixin` name (alternatively we could just reuse `Observed` but no strong opinions here.

packages/agents/src/mixin.ts
```diff
@@ -0,0 +1,648 @@
...
+}
+
+interface ObservedProperties {
+  _activeStreams: Set<SSEStreamingApi>;
+  _mcpConnections?: Map<string, MCPClientConnection>;
+  _fiberRouter?: Hono;
+}
+
+/**
+ * Mixin factory that adds Fiber capabilities to Agent classes
```
laulauland: can we have a better comment here.

maybe:

"Mixin factory that adds observability capabilities to Agent classes allowing them to be viewed in the Fiberplane agents playground"

PR #592, opened 2025-04-29
Update MCP support in the library and the example

laulauland:
In light of the changes to the agents SDK (https://github.com/cloudflare/agents/pull/217/files), `mcp` property for client connection manager is now hardcoded. Which means that we can simplify the property detection in the library.

It also updates the example to enable registering remote MCP connections (with authentication).


packages/agents/src/instrumentation.ts
```diff
@@ -169,9 +168,11 @@ function createAgentAdminRouter(agent: ObservedAgent) {
   });
 
   router.get("/agents/:namespace/:instance/admin/mcp", async (c) => {
-    agent._mcpConnections = detectMCPConnections(
-      agent as unknown as Record<string, unknown>,
-    );
+    // agent._mcpConnections = detectMCPConnections(
```
flenter: 👋

flenter (approved):

